PR #16, opened 2017-11-08
Add ENS registry address flag

izqui:
Adds support to passing a custom ENS Registry address as an option. This allows to use aragon-dev- cli in a dev network without a canonical ENS Registry

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -7,7 +7,7 @@ const cli = meow(`
     $ aragon-dev-cli <subcommand>
 
   Commands
-    init <name>                   Initialize a new Aragon module
+    init <name>                   Initialize a new Aragon module (e.g. test.aragonpm.eth)
     version <major|minor|patch>   Bump the module version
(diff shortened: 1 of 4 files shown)


onbjerg:
Can you use the `--registry` flag instead of `--ens`? `--registry` is only used in the `init` command, so just adjust it there.

izqui:
So you mean the same flag having different meanings on different commands? 

I think we should rename current `--registry` to `--apm-registry` and `--ens` to `--ens-registry`

onbjerg:
No, I meant only use `--registry` for ENS and remove it all together from the other commands

izqui:
Then the init command should take the full package name as its only argument: `aragon-dev-cli init test.aragonpm.eth`

onbjerg:
Yes

izqui:
Removed apm-registry flag, but keeping `--ens-registry` instead of just `--registry` for clarity

PR #20, opened 2017-11-14
Support scaffolding

### LINKED ISSUE #10: Add scaffolding
It should be possible to specify a template when running `aragon-dev-cli init` to scaffold a new module.

onbjerg:
Adds support for scaffolding in the `init` command (for example `aragon-dev-cli init [app ENS name] [template name, github repository name]`)

The command will clone a GitHub repository into a folder. It defaults to cloning aragon/aragon-react-boilerplate. The contents will be cloned to the "base name" of the application's ENS name, e.g. for foo.aragonpm.eth it would be cloned to a folder called foo.

Given we have multiple official boilerplates, we can alias them (for example, instead of typing `aragon/aragon-react-boilerplate`, simply typing `react` is good enough).

**Unsolved**
- We somehow need to update the `name` key of the `manifest.json` file in templates, but they might be in different locations

Closes #10

### DIFF
--- a/src/handle.js
+++ b/src/handle.js
@@ -13,6 +13,7 @@ const namehash = require('eth-ens-namehash')
 const inspector = require('solidity-inspector')
 const config = require('./config')()
 const pkgDir = require('pkg-dir')
+const clone = require('git-clone')
 
 const prettyKey = (key) => {
   let str = decamelize(key, ' ')
@@ -21,7 +22,7 @@ const prettyKey = (key) => {
 }
 
 const handlers = {
-  init ([name], flags) {
+  init ([name, template], flags) {
     if (!name) {
       reporter.fatal('No name specified')
     }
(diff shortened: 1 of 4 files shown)


src/handle.js
```diff
@@ -21,18 +22,27 @@ const prettyKey = (key) => {
...
+
+    // Determine template repository name
+    const aliasedTemplates = {
+      bare: 'aragon/aragon-bare-boilerplate',
+      react: 'aragon/aragon-react-boilerplate'
+    }
+    if (aliasedTemplates[template]) template = aliasedTemplates[template]
+
+    // Clone the template into the directory
+    // TODO: Somehow write name to `manifest.json` in template?
```
izqui: Need to appName on `module.json` also

src/handle.js
```diff
@@ -21,18 +22,27 @@ const prettyKey = (key) => {
...
+    if (!template) {
+      template = 'react'
+    }
+
+    // Determine template repository name
+    const aliasedTemplates = {
+      bare: 'aragon/aragon-bare-boilerplate',
+      react: 'aragon/aragon-react-boilerplate'
+    }
+    if (aliasedTemplates[template]) template = aliasedTemplates[template]
```
izqui: We should show an error (or at least warn the user) if an unknown template name is provided, instead of defaulting to the react boilerplate by default

onbjerg: Would 8f68ca5 suffice for that?

izqui: Yeah

PR #28, opened 2018-01-16
Add node version dependency

macor161:
I spent a few hours yesterday trying to run aragon-dev-cli when I realized I needed to update my node version from 8.1.4 to 8.9.1.

Just thought it would be nice for other devs who encounter the same problem to have a clear warning message when your node version is too old hehe.


onbjerg:
Yes, I've been meaning to do this. Thanks! 😊 

CI failing is my fault

PR #31, opened 2018-02-04
IPFS add: exclude paths

izqui:


onbjerg:
The way publishing works right now is that every file is copied to a temporary directory and published from there. Files *should* already be ignored by the copying mechanism, [here](https://github.com/aragon/aragon-dev-cli/blob/ed760429d587b6ed79bd6d289a4229fb294d1f7e/src/commands/publish.js#L122).

izqui:
Cool, should we close?

onbjerg:
Yes

PR #32, opened 2018-02-16
Fix errors when running init in clean directory

j0x0j:
I saw the repo at https://github.com/aragon/hack and tried to create a project in a clean directory and it failed. When I proceeded to debug I noticed the code in the npm registry is different from the master branch at aragon/aragon-dev-cli although the version is the same. So I cloned the project and tried to `init` but it failed again. After some digging around I found that since no `manifest.json` and no `module.json` files are found then the `init` command failed. I believe the intention there is to catch and show the error only in debug mode. When the error is caught it should be a no-op but since `argv.reporter` is undefined then the process throws on trying to call its `debug` method.

Hope this helps others get onboard in a more seamless manner.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,5 +52,7 @@
     "coveralls": "^3.0.0",
     "standard": "^10.0.3"
   },
-  "engines" : { "node" : ">=8.9.1" }
+  "engines": {
+    "node": ">=8.9.1"
+  }
 }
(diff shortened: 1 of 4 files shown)


onbjerg:
Sorry for getting back this late. This is no longer an issue, but thank you for your pull request!

PR #36, opened 2018-03-06
Rename bin from aragon-dev-cli to aragon

### LINKED ISSUE #35: Rename bin from `aragon-dev-cli` to just `aragon`
In case we ever make a CLI client, that could be called `dao`

```
aragon publish template.aragonpm.eth
dao new-org template.aragonpm.eth
```

izqui:
Close #35 

This will require several changes in documentation, leaving `aragon-dev-cli` as binary for a bit too

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Aragon command-line tools",
   "main": "src/cli.js",
   "bin": {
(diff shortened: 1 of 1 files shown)


onbjerg:
Great 🥇

PR #49, opened 2018-03-19
aragon grant

izqui:
New command for owner of a repo to allow other accounts to publish new versions to their repo


luisivan:
Nest grant?

onbjerg:
we automated it for you 😂

PR #69, opened 2018-05-11
Add more info to run command output

bingen:
Useful for debugging and interacting with DAO through console. Especially the ACL address to modify permissions.


onbjerg:
>App name: ${module.appName}
>App name hash: ${namehash.hash(module.appName)}
>App address: ${ctx.appAddress}

These are available in the dapp settings

>${chalk.bold('ACL address')}: ${ctx.aclAddress}

This is not going to be needed as I will add a more generalised `grant` and `revoke` command that finds this automatically. For end users they'll use the permissions app.

PR #92, opened 2018-05-17
Fix APM registry address

bingen:
Needs https://github.com/aragon/apm.js/pull/9


izqui:
But I like from this approach how you get it for the package concretely, will add that

bingen:
Nice, closing it then.

PR #95, opened 2018-05-19
Handle missing versions

bingen:
Show an error if any version is not found in `apm versions`.


src/commands/apm_cmds/versions.js
```diff
@@ -14,7 +14,11 @@ exports.handler = async function ({ reporter, module, bump, cwd, network, apm: a
   const versions = await APM(web3, apmOptions).getAllVersions(module.appName)
   reporter.info(`${module.appName} has ${versions.length} published versions`)
   versions.map((version) => {
-    reporter.success(`${version.version}: ${version.contractAddress} ${version.content.provider}:${version.content.location}`)
+    if (version === undefined) {
```
sohkai: Small JS hacks: `version == null` or `!version` may be better, depending on what the API guarantees.

In JS, `undefined == null`, and `!'' === true` (empty string casts to false).

Issue #70, opened 2018-05-15
Remove js-sha3 dependency


luisivan:
Other package already uses `sha3` instead of `js-sha3`, so we should remove the latter as a dependency

sohkai:
Done with #816.

PR #103, opened 2018-05-22
Remove js-sha3 dependency (#70)

Kyrrui:


luisivan:
Hey! Thanks for the PR! There's an issue, and it's that the CLI uses `js-sha3` in many places across its codebase. It would be trivial to replace, since we can have a `sha3(str)` function somewhere as a helper can mimicks the behaviour of `js-sha3`, but just using the compiled implementation `sha3` under the surface

Kyrrui:
@luisivan Ahh didn't realize that, I will go through and refactor. Is there some way to test the CLI so that I can run into this before creating a PR?

luisivan:
Right now there are no tests for the CLI, best we can do is a nice code review and try a few different commands (init, run, etc.)

luisivan:
They behave differently, `sha3` doesn't have a `keccak256` method and it works by updating the hash and then digesting, like this:
```javascript
let d = new SHA3.SHA3Hash();
d.update('foo');
d.digest('hex');
// => "1597842a..."
```
while in `js-sha3` you can just use `keccak256(data)`

izqui:
Is there any reason for changing the module use for `keccak256` hashing?

luisivan:
Mainly performance and not having two different implementations that do the same thing as dependencies

Kyrrui:
Sorry it's taken me so long to get back to this, still planning on doing it. Been busy

luisivan:
Hey @Kyrrui are you still planning on doing this?

Kyrrui:
@luisivan I apologize for the delay - I do plan on doing this, but for now it's on the back-burner for me. Feel free to hand off if it needs to be implemented sooner than later.

luisivan:
No rush for the moment!

luisivan:
Hey @Kyrrui! I take it that you won't take this issue anymore. I'll re-open it and leave unassigned for the moment!

PR #111, opened 2018-05-25
Use .ipfsignorefile for publish

bingen:
If it exists, otherwise stick to .gitignore.
Needed for https://github.com/aragon/aragon-react-boilerplate/pull/10


sohkai:
Our own apps don't currently have a `.gitignore` in their project roots, so I think this shouldn't break any of them.

bingen:
Cool. Besides, as long as you don't define this new `.ipfsignorefile`, everything should work the same way.

PR #112, opened 2018-05-25
Print mnemonic phrase when running the dev chain

sohkai:


izqui:
https://github.com/aragon/aragen/issues/16

PR #115, opened 2018-05-26
Add default ignore list for publishing to IPFS

sohkai:
Sets it to always ignore `node_modules` and `.git`, which I'm assuming nobody wants.

We could make it a default setting, but then if anyone wants to change it, they'd have to re-include the default list.

Seems to fix the issue with the `.ipfsignore` file, but not really sure why... (I think IPFS ignores the .gitignore dir anyway).


sohkai:
Closing, as this wasn't the issue with the publish command.

PR #141, opened 2018-06-19
Add call command

### LINKED ISSUE #122: Support for transaction pathing for performing actions
When sending transactions to apps from the CLI, some of them may require transaction pathing in order for the action to be eventually performed. 

Given that aragon.js is already being used in for the `dao` commands, some logic could be reused.

jvluso:
Creates a call command to path and call a function from an app. Resolves #122

### DIFF
--- a/src/commands/dao_cmds/utils/web3-utils.js
+++ b/src/commands/dao_cmds/utils/web3-utils.js
@@ -1,0 +1,12 @@
+import Web3 from 'web3'
+
+// Check address equality without checksums
+export function addressesEqual(first, second) {
+  first = first && first.toLowerCase()
+  second = second && second.toLowerCase()
+  return first === second
+}
+
+export function etherToWei(ether) {
+  return ether * 10e17
+}
--- a/README.md
+++ b/README.md
@@ -62,10 +62,10 @@ The local chain is started at `PORT`, which defaults to 8545.
 
 The publish step of the app works exactly as `aragon apm publish`, described below.
 
-#### `aragon apm publish` or `apm publish`
+#### `aragon apm publish`
 
 ```
-$ apm publish [CONTRACT_ADDRESS or CONTRACT_NAME] [--files PATTERN] [--ignore PATTERN] [--skip-confirm] [--only-artifacts]
+$ aragon apm publish [CONTRACT_ADDRESS or CONTRACT_NAME] [--files PATTERN] [--ignore PATTERN] [--skip-confirm] [--only-artifacts]
 ```
 
 Publishes a new (or the first!) version of your app.
@@ -80,30 +80,30 @@ Before publishing, your smart contract source code is scanned to generate an art
 
 If `--skip-confirm` is specified the command will not wait for the transaction to receive confirmations.
 
-#### `aragon apm version` or `apm version`
+#### `aragon apm version`
 
 ```
-$ apm version <BUMP>
+$ aragon apm version <BUMP>
 ```
 
 Bumps the version of your Aragon app, where a valid `BUMP` is either *major*, *minor* or *patch*.
 
 Note that you are only allowed to release a version with a new smart contract address if the bump specified is major.
 
-#### `aragon apm versions` or `apm versions`
+#### `aragon apm versions`
 
 ```
-$ apm versions
+$ aragon apm versions
 ```
 
 View a list of published versions for the app in the current directory.
 
 ### DAOs
 
-#### `aragon apm grant` or `apm grant`
+#### `aragon apm grant`
 
 ```
-$ apm grant <ADDRESS> [--skip-confirm]
+$ aragon apm grant <ADDRESS> [--skip-confirm]
 ```
 
 Grants permission for `ADDRESS` to interact with the APM repository of the app in the current directory.
--- a/aliases/apm
+++ b/aliases/apm
@@ -1,2 +1,0 @@
-#!/bin/sh
-aragon apm $@
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "main": "dist/cli.js",
   "bin": {
     "aragon": "./dist/cli.js",
-    "apm": "./aliases/apm",
     "dao": "./aliases/dao"
   },
   "scripts": {
--- a/src/cli.js
+++ b/src/cli.js
@@ -88,7 +88,11 @@ cmd.option('network', {
     }
     let provider
     if (truffleNetwork.provider) {
-      provider = truffleNetwork.provider
+      if (typeof truffleNetwork.provider === 'function') {
+        provider = truffleNetwork.provider()
+      } else {
+        provider = truffleNetwork.provider
+      }
     } else if (truffleNetwork.host && truffleNetwork.port) {
       provider = new Web3.providers.WebsocketProvider(`ws://${truffleNetwork.host}:${truffleNetwork.port}`)
     } else {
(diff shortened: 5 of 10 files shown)


izqui:
Thanks for the contribution! Will review soon.

sohkai:
Pinging @Quazia to review as well :).

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+const knownRoles = rolesForApps()
+const ANY_ENTITY = '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF'
+const ANY_ENTITY_TEXT = 'Any entity'
+
+const path = require('path')
+let knownApps
+
+exports.command = 'call <dao> <proxy-address> <fn> [fn-args]'
+
+exports.describe = 'Inspect permissions in a DAO'
```
izqui: Should describe this command: 'Executes a call in an app of a DAO'

src/commands/dao_cmds/utils/web3-utils.js
```diff
@@ -0,0 +1,27 @@
...
+// Check address equality without checksums
+export function addressesEqual(first, second) {
+  first = first && first.toLowerCase()
+  second = second && second.toLowerCase()
+  return first === second
+}
+
+// Cache web3 instances used in the app
+const cache = new WeakMap()
+export function getWeb3(provider) {
```
izqui: This function is not used, nor I think this caching is needed.

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+exports.handler = async function ({ reporter, dao, apm, network, proxyAddress, fn, fnArgs }) {
+  knownApps = listApps()
+  const web3 = await ensureWeb3(network)
+
+  const tasks = new TaskList([
+    {
+      title: 'Generating transaction',
+      task: async (ctx, task) => {
+        task.output = `Fetching dao at ${dao}...`
+
```
izqui: One line of whitespace is enough here

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+
+        if (!ctx.accounts) {
+          ctx.accounts = await web3.eth.getAccounts()
+        }
+
+        return new Promise((resolve, reject) => {
+
```
izqui: No whitespace is needed here, would remove both new lines

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+const colors = require('colors')
+
+const knownRoles = rolesForApps()
+const ANY_ENTITY = '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF'
+const ANY_ENTITY_TEXT = 'Any entity'
+
+const path = require('path')
+let knownApps
+
+exports.command = 'call <dao> <proxy-address> <fn> [fn-args]'
```
izqui: I think we should rename `call` to `exec`, so it doesn't conflict with the concept of a call in web3, which doesn't modify state.

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+            })
+        })
+      }
+    },
+    {
+      title: `calling ${fn}`,
+      task: async (ctx, task) => {
+        task.output = `Waiting for response...`
+        ctx.transactionPath[0].gas = await web3.eth.estimateGas(ctx.transactionPath[0])
+        ctx.transactionPath[0].gas = ctx.transactionPath[0].gas *2 // need better estimations
```
izqui: Would rewrite to:

```js
let tx = ctx.transactionPath[0] // TODO: Support choosing between possible transaction paths

if (!tx) {
  throw new Error('Cannot find transaction path for executing action')
}

const estimatedGas = await web3.eth.estimateGas(ctx.transactionPath[0])
tx.gas = parseInt(GAS_ESTIMATE_ERROR_FACTOR * estimatedGas)
```

`const GAS_ESTIMATE_ERROR_FACTOR` should be defined near `ANY_ENTITY`

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+            .catch(err => {
+              reporter.error('Error inspecting DAO')
+              reporter.debug(err)
+              process.exit(1)
+            })
+        })
+      }
+    },
+    {
+      title: `calling ${fn}`,
```
izqui: Capitalize: `Executing ${fn} in ${proxyAddress}`

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+      task: async (ctx, task) => {
+        task.output = `Waiting for response...`
+        ctx.transactionPath[0].gas = await web3.eth.estimateGas(ctx.transactionPath[0])
+        ctx.transactionPath[0].gas = ctx.transactionPath[0].gas *2 // need better estimations
+        return new Promise((resolve, reject) => {
+          web3.eth.sendTransaction(ctx.transactionPath[0],(err,res) => {
+            if(err) reject(err)
+            ctx.res=res
```
izqui: Style: `ctx.res = res`

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+    {
+      title: `calling ${fn}`,
+      task: async (ctx, task) => {
+        task.output = `Waiting for response...`
+        ctx.transactionPath[0].gas = await web3.eth.estimateGas(ctx.transactionPath[0])
+        ctx.transactionPath[0].gas = ctx.transactionPath[0].gas *2 // need better estimations
+        return new Promise((resolve, reject) => {
+          web3.eth.sendTransaction(ctx.transactionPath[0],(err,res) => {
+            if(err) reject(err)
```
izqui: Should return here, otherwise it will also call `resolve()`:

```js
if (err) {
  reject(err)
  return
}
```

src/commands/dao_cmds/call.js
```diff
@@ -0,0 +1,113 @@
...
+              ctx.apps = apps
+              if (ctx.wrapper) {
+                ctx.transactionPath = await ctx.wrapper.getTransactionPath(proxyAddress, fn, fnArgs)
+                resolve()
+              }
+            },
+            onDaoAddress: addr => ctx.daoAddress = addr,
+            onError: err => reject(err)
+          })
+            .then(async wrapper => {
```
izqui: Line 76-87 should be indented one level down.

src/commands/dao_cmds/utils/web3-utils.js
```diff
@@ -0,0 +1,27 @@
...
+  if (cache.has(provider)) {
+    return cache.get(provider)
+  }
+  const web3 = new Web3(provider)
+  cache.set(provider, web3)
+  return web3
+}
+
+export function weiToEther(wei) {
```
izqui: These functions are not used either, and it is better to use `web3.fromWei` as they handle BigNumber

izqui:
@jvluso did you see my review comments? When do you think you will have time to address them? We are looking forward to merging it :)

jvluso:
@izqui I did see it. I've spent this week incorporating and getting https://recoblix.com/ up and running. Hopefully next week I can get back to writing software.

izqui:
@jvluso can you please rebase your actual changes with the current master branch? There are a ton of unrelated commits in this PR, some of which aren't even in the master branch right now.

PR #145, opened 2018-06-20
Pass contract name specified in publish to deploy

### LINKED ISSUE #121: Fix yargs complaining about lack of positional arguments with defaults
Even when there are defaults for optional default arguments, yargs will complain that a required parameter hasn't been set. 

This happens in `aragon apm publish` and `aragon apm version` when some flags are passed (like network)

### LINKED ISSUE #138: Fix `apm publish X` always publishing contract in arapp.json rather than X

### LINKED ISSUE #142: `aragon --version` and `aragon --help` shouldn't show an error when not run inside an aragon app project
```
$ aragon --version
Error: This directory is not an Aragon project
[ redacted full error message]
4.1.2
```
It should just print `4.1.2` without the error.

izqui:
Close #138 

While working on this issue, I run into an old issue regarding yargs re-coercing the network flag. The observed behavior was: when using a command with an optional positional argument, if no argument is passed then it was fine, but if the argument is specified, a bug in yargs would try to coerce the network flag with the default 'development' flag, which in case it wasn't present it would throw an error (but continue execution), and make the positional argument undefined, even if explicitly defined. 

So this also finally fixes #121

(Also fixes #142, a stupid typo that caused a bug not detected until now https://github.com/aragon/aragon-cli/pull/145/files#diff-d9d9214113dea1c364abd4ba3c70be11R70)

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -65,17 +65,17 @@ cmd.option('network', {
     }
 
     // Catch commands that dont require network and return
-    const skipNetworkSubcommands = ['version', 'new']
+    const skipNetworkSubcommands = new Set(['version']) // 'aragon apm version'
     if (process.argv.length >= 4) {
-      if (skipNetworkSubcommands.indexOf(process.argv[3]) >= 0) {
+      if (skipNetworkSubcommands.has(process.argv[3])) {
         return {}
       }
     }
 
-    const skipNetworkCommands = ['init', 'devchain', 'ipfs']
+    const skipNetworkCommands = new Set(['init', 'devchain', 'ipfs'])
 
     if (process.argv.length >= 3) {
-      if (skipNetworkCommands.indexOf(process.argv[2]) >= 0) {
+      if (skipNetworkCommands.has(process.argv[2])) {
         return {}
       }
     }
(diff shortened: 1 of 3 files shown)


src/cli.js
```diff
@@ -45,15 +45,29 @@ cmd.option('cwd', {
...
   coerce: (network) => {
+    if (cachedNetwork) {
+      return cachedNetwork
+    }
+
     // Catch commands that dont require network and return
     const skipNetworkSubcommands = ['version', 'new']
     if (process.argv.length >= 4) {
-      if (skipNetworkSubcommands.indexOf(process.argv[3]) >= 'version') {
+      if (skipNetworkSubcommands.indexOf(process.argv[3]) >= 0) {
```
sohkai: Another way to do this is either with an object or, more preferably, a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), which would've avoided the typo.

sohkai:
Another thing to consider is also updating the yargs dependency; it looks like [`12.0.0`](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1200-2018-06-26) has a lot of fixes (can't tell if they're relevant to the coercing bug though).

PR #155, opened 2018-07-10
janky network id fix

jtremback:


sohkai:
@izqui Maybe we can just use a hardcoded network id, e.g. 15?

src/commands/devchain.js
```diff
@@ -54,7 +54,8 @@ exports.task = async function ({ port = 8545, reset = false, showAccounts = 2 })
       const server = ganache.server({
         gasLimit: BLOCK_GAS_LIMIT,
         mnemonic: MNEMONIC,
-        db_path: snapshotPath
+        db_path: snapshotPath,
+        network_id: Math.floor(Math.random() * 1000)
```
izqui: As @sohkai points out, the `networkId` that is used to generate the snapshot is `15`. We should keep using 15 to avoid potential issues.

jtremback:
Sorry guys, shouldn't have made this a PR... I'm not sure this is a good fix.

PR #157, opened 2018-07-11
Fix opening IPFS documentation URL

mnaamani:
Import of `opn` module was missing.

Fixes call to `opn('https://ipfs.io/docs/install')` in:

https://github.com/aragon/aragon-cli/blob/bacfdf97ca8c91d1f2cb9ee437e46c3964cbe657/src/commands/ipfs.js#L26


sohkai:
Thanks @mnaamani, and good catch!

Issue #93, opened 2018-05-18
Add functional tests


izqui:
Given the increased complexity with new features, modifying a command can have side-effects in other commands that rely on it. 

My proposal is that the initial test suite for the CLI should have a bunch of tests that interact with the CLI directly. Then each test case would pass if, it exits without an error and the side-effects of the command are what was expected (e.g. check the chain to see if a DAO or APM package is at the correct state, or check artifacts against the expected state)

PascalPrecht:
I looked into this and I think it'd make sense to decouple the command tasks from the actual command implementations. 

For example, if we'd wanna write a test for `aragon init`, it comes with the side effect that a template repo is being cloned and npm dependencies are being installed. This can slow down the testing experience, so I think it'd be nice if we could test the tasks in an isolated fashion (not necessarily through the CLI). For that however, the tasks need to be extracted into their own module (or at least exported separately).

Would you be okay with that?

izqui:
Do you mean the individual tasks or the command tasks? Most of the commands export their `TaskList` at `command.task`, so they can already be tested without testing the CLI itself.

It would be great to see how what you have in mind would look like, just an example of how a small command would change to allow this would be great

PascalPrecht:
> Most of the commands export their TaskList at command.task, so they can already be tested without testing the CLI itself.

Exactly. Even that however would mean, one either tests the whole task list as a whole or nothing, unless one can access specific tasks from that list (not super familiar with the TaskList API yet).

So in my head a command could look like this:

```
exports.handler = (...) => {
  const tasks = new TaskList([
    {
      ...
      task: async someTaskFunction
    }
  ])

  tasks.run(...)
}
```

`someTaskFunction` would be defined somewhere (probably imported). That way `someTaskFunction` can be tested without even knowing its part of a CLI. Obviously one can break it down further into smaller bits and pieces as different tasks may have different "sub" tasks.

Basically where I'm trying to get is a setup where for example the `init` tasks can be tested, without necessarily having to test (and therefore execute) the tasks of `init` that introduce the "negative" side effects (like npm install, git clone etc).

This is also not a solution to the problem, more a "how about we go this way for now, instead of not having any tests at all"

Hope this makes sense!

sohkai:
I'm definitely a fan of how [pando's structured their cli code](https://github.com/ryhope/pando/tree/master/packages/cli/src/commands), that sounds similar to what you're suggesting @PascalPrecht.

[...]

PR #162, opened 2018-08-06
fix: don't use yarn for aragon init

### LINKED ISSUE #129: Don't use yarn for aragon init
Web3.js has issues being installed by yarn (lol 🤦‍♂️). We should disable the yarn detection when installing modules.

PascalPrecht:
As mentioned in #129, Web3.js has issues with being installed using yarn.
This commit defaults to the npm as package manager to install dependencies
in `aragon init`.

Notice that this commit can be reverted once the issues with yarn in Web3.js
have been resolved.

Closes #129

### DIFF
--- a/src/commands/apm_cmds/publish.js
+++ b/src/commands/apm_cmds/publish.js
@@ -312,7 +312,7 @@ exports.task = function ({
           return
         }
 
-        const bin = await getNodePackageManager()
+        const bin = getNodePackageManager()
         const buildTask = execa(bin, ['run', 'build'])
         buildTask.stdout.on('data', (log) => {
           if (!log) return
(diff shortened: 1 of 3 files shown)


PascalPrecht:
Yay!

@izqui I wanted to provide a test as well as I've noticed that CI is failing due to missing test files. When I started working on setting this up, I've noticed there'd be no "easy" way to test this change. It'd be either an e2e test that would work in a sandboxed environment, or we'd try to mock `installDeps()`. However, the latter seems rather hard to do without DI.

Where can I find information about testing plans for CLI?

izqui:
@PascalPrecht There is an open issue for setting up the testing suite for the project: https://github.com/aragon/aragon-cli/issues/93. Given the amount of improvements we can still do to the CLI, figuring out testing is not the top priority at the moment, but setting up some basic functional tests (so we can have peace of mind that new changes don't break any of the working user stories) would be great. It would be an amazing contribution :)

PR #164, opened 2018-08-07
fix(commands/init): throw meaningful error if project folder already exists

### LINKED ISSUE #163: Improve UX of init command when project folder already exists
When running the `init` command with a project name that happens to be used by another folder in the current working directory, aragon CLI errors out in a user unfriendly fashion:

```
~/projects/machinelabs
❯ aragon init machinelabs.aragonpm.eth
 ✖ Clone template
   → 'git clone' failed with status 128
   Preparing template
   Install package dependencies
 ✖ 'git clone' failed with status 128
```

Since there's already a folder `machinelabs` (which also happens to be a git repo), `init` fails but it's not very obvious at this point, why it fails.

I guess it'd be better if CLI would check whether the project (or a folder with that name) already exists, and if that's the case, either provide a meaningful error message, or let the user decide to wipeout/override the original project folder.

PascalPrecht:
Closes #163

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -47,6 +47,16 @@ exports.handler = function ({ reporter, name, template }) {
   const basename = name.split('.')[0]
   const tasks = new TaskList([
     {
+      title: 'Checking project existence',
+      task: async (ctx, task) => {
+        const projectPath = path.resolve(process.cwd(), basename)
+        const exists = await fs.pathExists(projectPath)
+        if (exists) {
+          throw new Error(`Couldn't initialize project. Project with name ${basename} already exists in ${projectPath}. Use different <name> or rename existing project folder.`)
+        }
+      }
+    },
+    {
       title: 'Clone template',
       task: async (ctx, task) => {
         task.output = `Cloning ${template} into ${basename}...`
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -62,7 +62,7 @@ exports.handler = function ({ reporter, name, template }) {
         task.output = `Cloning ${template} into ${basename}...`
 
         const repo = await clone(template, basename, { shallow: true })
-        console.log(`Template cloned to ${basename}`)
+        // console.log(`Template cloned to ${basename}`)
       }
     },
     {


src/commands/init.js
```diff
@@ -46,6 +46,16 @@ exports.handler = function ({ reporter, name, template }) {
   
   const basename = name.split('.')[0]
   const tasks = new TaskList([
+    {
+      title: 'Checking project existence',
+      task: async (ctx, task) => {
```
PascalPrecht: `ctx` and `task` aren't really used here, but have been provided in other tasks that didn't use them as well. I just stuck to your style. Let me know if you want me to remove those.

src/commands/init.js
```diff
@@ -46,6 +46,16 @@ exports.handler = function ({ reporter, name, template }) {
...
+        const projectPath = path.resolve(process.cwd(), basename)
+        const exists = await fs.pathExists(projectPath)
+        if (exists) {
+          throw new Error(`Couldn't initialize project. Project with name ${basename} already exists in ${projectPath}. Use different <name> or rename existing project folder.`)
```
PascalPrecht: First stab on meaningful error message. Happy to add better wording here if anyone has strong opinions :)

PascalPrecht:
I just realized I've used a commit message convention that's not really applied in this project (sorry habit). If you want me to change it, please lemme know! :)

luisivan:
I actually think we should start using that commit message convention, so no problem @PascalPrecht :)

PascalPrecht:
Ha sweet. Comes with the nice side effect that we can generate changelogs automatically from it. In case you're interested [this](https://github.com/machinelabs/machinelabs/blob/master/CHANGELOG.md) is what that looks like.

luisivan:
That's amazing. We should adopt this @sohkai

PascalPrecht:
For more info, this [section](https://github.com/machinelabs/machinelabs/blob/master/CONTRIBUTING.md#commit) explains the conventions.

src/commands/init.js
izqui: Lets keep them for consistency :)

src/commands/init.js
izqui: LGTM!

src/commands/init.js
```diff
@@ -46,13 +46,23 @@ exports.handler = function ({ reporter, name, template }) {
...
+      }
+    },
     {
       title: 'Clone template',
       task: async (ctx, task) => {
         task.output = `Cloning ${template} into ${basename}...`
 
         const repo = await clone(template, basename, { shallow: true })
-        console.log(`Template cloned to ${basename}`)
+        // console.log(`Template cloned to ${basename}`)
```
PascalPrecht: How would you feel about removing it entirely instead of leaving a comment here @izqui ? :)

Also, would you mind enlighten me how having this console.log causes multiple outputs of a totally different task?

izqui: Your are absolutely right, forgot to remove it after testing. Fixed: https://github.com/aragon/aragon-cli/pull/165

`TaskList` basically hijacks the console output and if you try to print in a different way, the output gets fucked up.

PascalPrecht: > TaskList basically hijacks the console output and if you try to print in a different way, the output gets fucked up.

Wow.. I had no idea. Thanks!

Issue #98, opened 2018-05-21
Create 'aragon start' command using wrapper's 'npm run start:dev'


izqui:
- [ ] Implement `aragon start [git-ref]`
- [ ] `aragon run` should use `aragon start` to start wrapper

See https://github.com/aragon/aragon/pull/241

galactusss:
Hi I'm trying to tackle this issue. I been researching and I have some questions:

1. I need to create a totally new aragon command (like aragon run, aragon init, etc) for start?
2. The second task, it means to add the new aragon start command as a new task in aragon run TaskList?

sohkai:
@galactusss Yup :).

It's likely to involve making the [`Open DAO`](https://github.com/aragon/aragon-cli/blob/master/src/commands/run.js#L247) tasks its own command and then calling it in `aragon run` like [how we do a few of the other commands](https://github.com/aragon/aragon-cli/blob/master/src/commands/run.js#L161).

galactusss:
Thank you for your answer @sohkai it helped me to keep forward. I kind of start understanding yargs. 

I would like to have some guide about the positional argument [git-ref] My guess is that 'git-ref' is the git repo of the wrapper, but I'm not sure.

sohkai:
`[git-ref]` would be a commit hash, branch name, or tag name. Basically something that can be used with `git checkout`.

izqui:
Actually `aragon start` should use the new flags `aragon run` accepts for configuring how the client is run: `--client-version` (what was referred here as `[git-ref]`), `--client-port` and `--client-path`

0x6431346e:
Perhaps we could name this `aragon start-client` ?

PR #168, opened 2018-08-10
refactor: move init command functions into lib and introduce tests

PascalPrecht:
@izqui @sohkai here's a first POC that shows what I was aiming for. Notice how the individual tasks have been extracted from the TaskList, so they can be imported individually and therefore be tested in isolation.

I added some inline comments with some additional thoughts as well. 

This is definitely not the most beautiful architecture, but I think it takes us one step further towards a testable code base.

Let me know what you think.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -6,6 +6,7 @@ const path = require('path')
 const fs = require('fs-extra')
 const { installDeps } = require('../util')
 const defaultAPMName = require('../helpers/default-apm')
+import { checkProjectExists, prepareTemplate } from '../lib/init';
 
 exports.command = 'init <name> [template]'
 
@@ -43,52 +44,32 @@ exports.builder = (yargs) => {
 
 exports.handler = function ({ reporter, name, template }) {
   name = defaultAPMName(name)
-  
   const basename = name.split('.')[0]
+
   const tasks = new TaskList([
     {
-      title: 'Checking project existence',
+      title: 'Preparing initialization',
       task: async (ctx, task) => {
-        const projectPath = path.resolve(process.cwd(), basename)
-        const exists = await fs.pathExists(projectPath)
-        if (exists) {
-          throw new Error(`Couldn't initialize project. Project with name ${basename} already exists in ${projectPath}. Use different <name> or rename existing project folder.`)
-        }
+        task.output = 'Checking if project folder already exists...'
+        await checkProjectExists(basename)
       }
     },
     {
-      title: 'Clone template',
+      title: 'Cloning app template',
       task: async (ctx, task) => {
         task.output = `Cloning ${template} into ${basename}...`
-
-        const repo = await clone(template, basename, { shallow: true })
+        await clone(template, basename, { shallow: true })
       }
     },
     {
       title: 'Preparing template',
       task: async (ctx, task) => {
-        // Set `appName` in arapp
-        const arappPath = path.resolve(
-          basename,
-          'arapp.json'
-        )
-        const arapp = await fs.readJson(arappPath)
-        arapp.appName = name
-
-        // Delete .git folder
-        const gitFolderPath = path.resolve(
-          basename,
-          '.git'
-        )
-
-        return Promise.all([
-          fs.writeJson(arappPath, arapp, { spaces: 2 }),
-          fs.remove(gitFolderPath)
-        ])
+        task.output = 'Initiliazing arapp.json and removing Git repository';
+        await prepareTemplate(basename, name)
       }
     },
     {
-      title: 'Install package dependencies',
+      title: 'Installing package dependencies',
       task: async (ctx, task) => (await installDeps(basename, task)),
     }
   ])
--- a/package.json
+++ b/package.json
@@ -88,6 +88,9 @@
   "ava": {
     "require": [
       "@babel/register"
+    ],
+    "files": [
+      "test/**/*.js"
     ]
   }
 }
--- a/src/lib/init/check-project-exists.js
+++ b/src/lib/init/check-project-exists.js
@@ -1,0 +1,11 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+export async function checkProjectExists(basename) {
+  const projectPath = path.resolve(process.cwd(), basename)
+  const exists = await fs.pathExists(projectPath)
+
+  if (exists) {
+    throw new Error(`Couldn't initialize project. Project with name ${basename} already exists in ${projectPath}. Use different <name> or rename existing project folder.`)
+  }
+}
--- a/src/lib/init/index.js
+++ b/src/lib/init/index.js
@@ -1,0 +1,2 @@
+export * from './check-project-exists';
+export * from './prepare-template';
--- a/src/lib/init/prepare-template.js
+++ b/src/lib/init/prepare-template.js
@@ -1,0 +1,16 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+export async function prepareTemplate(basename, appName) {
+  const arappPath = path.resolve(basename, 'arapp.json')
+  const arapp = await fs.readJson(arappPath)
+
+  arapp.appName = appName
+
+  const gitFolderPath = path.resolve(basename, '.git')
+
+  return Promise.all([
+    fs.writeJson(arappPath, arapp, { spaces: 2 }),
+    fs.remove(gitFolderPath)
+  ])
+}
--- a/test/commands/init.js
+++ b/test/commands/init.js
@@ -1,0 +1,43 @@
+import test from 'ava';
+import fs from 'fs-extra';
+
+import { checkProjectExists, prepareTemplate } from '../../src/lib/init';
+
+const projectPath = './.tmp/aragon-app'
+
+test.beforeEach(t => {
+  fs.ensureDirSync(projectPath)
+})
+
+test.afterEach(t => {
+  fs.removeSync(projectPath)
+})
+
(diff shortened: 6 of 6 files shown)


src/commands/init.js
```diff
@@ -7,6 +7,64 @@ const fs = require('fs-extra')
 const { installDeps } = require('../util')
 const defaultAPMName = require('../helpers/default-apm')
 
+
+const handlerTasks = {
+  checkProjectExistence: (options) => {
```
PascalPrecht: `handlerTasks` is exported below so tasks can be imported individually when testing.

Also, each handler task receives options. That's because a command's `handler` might introduce some additional logic and properties that may be needed within handler tasks.

src/commands/init.js
```diff
@@ -41,56 +99,17 @@ exports.builder = (yargs) => {
...
-      }
-    },
-    {
-      title: 'Install package dependencies',
-      task: async (ctx, task) => (await installDeps(basename, task)),
-    }
+    handlerTasks.checkProjectExistence({ basename }),
+    handlerTasks.cloneTemplate({ template, basename }),
+    handlerTasks.prepareTemplate({ name, basename }),
+    handlerTasks.installDependencies({ basename })
```
PascalPrecht: I'd even take this one step further and do it similar to how Pando is doing it (as @sohkai mentioned), where there could be a `AragonCLI` library, that comes with all the tasks, which we can then use in CLI wrapper commands.

The library would be the core that'd be tested, while the CLI wrapper commands could be used for e2e tests.

PascalPrecht: For reference: https://github.com/ryhope/pando/blob/master/packages/cli/src/commands/init/index.ts#L2

src/commands/init.js
izqui: I think in general `options` should be replaced by the actual option values that this particular task needs, in this case: `options` -> `{ basename }`.

src/commands/init.js
izqui: I like that idea actually. Also I think the task title shouldn't be at library level (they are the CLI's UI), as it kind of is right now by being provided as a property to the task object (due to how TaskList works)

src/commands/init.js
```diff
@@ -41,56 +99,17 @@ exports.builder = (yargs) => {
...
-          fs.writeJson(arappPath, arapp, { spaces: 2 }),
-          fs.remove(gitFolderPath)
-        ])
-      }
-    },
-    {
-      title: 'Install package dependencies',
-      task: async (ctx, task) => (await installDeps(basename, task)),
-    }
+    handlerTasks.checkProjectExistence({ basename }),
```
izqui: Just a note: in some commands, context from one task to another is passed by writing to the `ctx` object. In order to test individual tasks, the context will need to be properly set. Example: https://github.com/aragon/aragon-cli/blob/master/src/commands/apm_cmds/publish.js#L217

src/commands/init.js
PascalPrecht: Yes, I agree with that. We can roll that back. In fact, right now I'd vote for introducing a `lib` folder with all the functions that will be consumed by tasklist (and therefor CLI wrappers).

Once that is done, we can think about how to move away from Tasklist (if this is still something that should happen).

src/commands/init.js
PascalPrecht: 👍

src/commands/init.js
PascalPrecht: Yes makes sense. It probably makes sense to find a general API that covers all possible cases, so that, whenever a new CLI functions is introduced, they expect the same parameters etc. Will have to look at other tasks to see what exactly would work. But we can definitely just go with destruction patterns for now.

PascalPrecht:
> I would say progressively moving away from TaskList could be a good decision

Depends.. I assume TaskList (listr) has been introduced for a good reason. So maybe we can figure out first, what TaskList adds to the able and then decide how to move on with it?

izqui:
I think it was introduced mostly for UI reasons when working on the `aragon run` command, that has a bunch of steps and sub-tasks. Since that it has been extended to all commands for consistency's sake. The limited flexibility for outputting to the console while tasks are running is a major pain point.

My comment about removing it all-together is a bit extreme, but I think it shouldn't influence how the CLI is architected too much.

PascalPrecht:
Gotcha, thanks!

src/commands/init.js
```diff
@@ -6,6 +6,8 @@ const path = require('path')
 const fs = require('fs-extra')
 const { installDeps } = require('../util')
 const defaultAPMName = require('../helpers/default-apm')
+import { checkProjectExists } from '../lib/check-project-exists';
+import { prepareTemplate } from '../lib/prepare-template';
```
PascalPrecht: As discussed, we're introducing a `lib` folder here that would have all the actual functions etc. that can be used to compose CLI commands.

Also notice that I'm using `import` syntax here. This doesn't look very nice as the project so far seems to use commonjs syntax. I'm happy to change that to commonjs, however, I'd personally vote for rather slowly moving to ECMAScript module syntax as this is an official standard (by now).

If you agree with that, we can decide whether we change those imports in one go, or progressively as we go along (also here my personally vote: better progressively and keep shipping at the cost of having some temporary inconsistencies)

src/commands/init.js
```diff
@@ -43,52 +45,32 @@ exports.builder = (yargs) => {
...
-          fs.writeJson(arappPath, arapp, { spaces: 2 }),
-          fs.remove(gitFolderPath)
-        ])
+        task.output = 'Initiliazing arapp.json and removing Git repository';
+        await prepareTemplate(basename, name)
       }
     },
     {
-      title: 'Install package dependencies',
+      title: 'Installing package dependencies',
```
PascalPrecht: Basically when back to having the tasks defined inlined, as the actual functionality is moved into lib modules now. That way, we can easily move away from tasklist at some point, without having to worry about touch test imports.

src/lib/prepare-template.js
```diff
@@ -0,0 +1,16 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+export async function prepareTemplate(basename, appName) {
```
PascalPrecht: @izqui As you can see, I didn't go with the destruction syntax here any more. The main reason really is that, we keep composing the tasks within the handler where we have full control over what parameters are being passed to library functions.

Happy to go with `options` object with destruction syntax again, if you have strong feelings about that.

PascalPrecht:
Alright, updated the PR and added some inline comments. I basically rolled back to having inline tasks but moving the core functionalities into lib modules. 

The nice thing about that is that we can now start writing tests for those library functions, without the whole tasklist ceremony. In fact, we also don't have to worry about tasks that use third party code we don't have control over (e.g. `clone()` should be tested by `git-clone`'s author).

I think this could be a reasonable start to move the CLI structure in a more testable direction.

PTAL :)

src/commands/init.js
izqui: Looks good. The `lib` directory may end up having a ton of files if we extract all individual tasks. I propose we create a directory for every command (the ones you have created now would go under `lib/init/`).

If we do that, how would you feel about having an index file per command in the lib dir and export all the different tasks?

```
import { prepareTemplate, checkProjectExists } from '../lib/init'
```

For the import/exports, I agree on changing as we go.

src/commands/init.js
izqui: Perfect

src/lib/prepare-template.js
izqui: 👍 I think explicitly passing the parameters each function needs is better than the options object.

src/commands/init.js
PascalPrecht: > I propose we create a directory for every command (the ones you have created now would go under lib/init/).

We can certainly do that, just keep in mind that there might be lib functions that will be used across tasks, or maybe even other lib functions. However, lib functions that exists only for a dedicated command, then definitely go in corresponding sub folders.

> If we do that, how would you feel about having an index file per command in the lib dir and export all the different tasks?

Sure, we can do that. No strong feelings about that. :)

src/lib/prepare-template.js
PascalPrecht: Wonderful!

PascalPrecht:
> The only issue with this could be executing sub-TaskLists (which we use quite a bit for complex commands), that automatically inherit the context and can write to it (for example, the publish command uses the deploy command to deploy a contract and then reads from the context the address where the contract was deployed to here). We need to think how these cases would look like: the logic of task is just executing another task.

Yes. I haven't actually paid attention to those yet because I first wanted to make a very concrete example of what the proposed change would look like. 

I'd propose, I update this PR once more with the changes you've requested (introducing sub folders in `lib`) and also adding first tests and then we can take it from there and work our way through command by command.

PascalPrecht:
@izqui this is now updated and introduces first couple of tests.

0x6431346e:
This looks great :+1: 
@izqui should we merge this?

izqui:
Agreed! Feel free to merge @0x6431346e after testing it :)

0x6431346e:
> Agreed! Feel free to merge @0x6431346e after testing it :)

Done. Tested by building and using `npm link` to init a new project :100:

PascalPrecht:
Yay! Happy this has landed!

izqui:
Thanks for this amazing contribution @PascalPrecht!!

src/commands/init.js
galactusss: I'm working on #98, is ok to refactor my implementation of the new command aragon start?

PR #170, opened 2018-08-10
Improve experience around devchain resets

### LINKED ISSUE #156: Metamask nonce issue
I was running into this issue: https://github.com/MetaMask/metamask-extension/issues/1999

I fixed it with this janky hack: https://github.com/aragon/aragon-cli/pull/155/commits/e2509c221ede7596b2f911d7dac9ca623d897585

Is there a better long term solution?

izqui:
Close #156 

Starting the devchain with a different networkID every time doesn't appear to have any side effects, and it makes Metamask detect that the network has changed and prevents it from using a previously cached nonce when sending transactions. Sometimes you need to switch to another network and then switch back to the local dev network, but then it works without any problem.

Added some warning messages when resetting the chain, prompting users to do this so they don't have problems with it.

### DIFF
--- a/src/commands/devchain.js
+++ b/src/commands/devchain.js
@@ -108,7 +108,7 @@ exports.printMnemonic = (reporter, mnemonic) => {
 
 exports.printResetNotice = (reporter, reset) => {
   if (reset) {
-    reporter.warning(`The devchain was reset, some steps need to be done to provent issues:
+    reporter.warning(`The devchain was reset, some steps need to be done to prevent issues:
     - Reset the application cache in Aragon Core by going to Settings > Troubleshooting.
     - If using Metamask: switch to a different network, and then switch back to the 'Private Network' (this will clear the nonce cache and prevent errors when sending transactions)  
   `)
--- a/src/commands/devchain.js
+++ b/src/commands/devchain.js
@@ -52,10 +52,12 @@ exports.task = async function ({ port = 8545, reset = false, showAccounts = 2 })
     title: 'Starting a local chain from snapshot',
     task: async (ctx, task) => {
       const server = ganache.server({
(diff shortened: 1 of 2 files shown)


src/commands/devchain.js
```diff
@@ -88,8 +90,10 @@ exports.task = async function ({ port = 8545, reset = false, showAccounts = 2 })
 }
 
 exports.printAccounts = (reporter, privateKeys) => {
+  const firstAccountComment = '(this account is used to deploy DAOs, it has more permissions)'
+
   const formattedAccounts = privateKeys.map(({ address, key }, i) => 
-    chalk.bold(`Address #${i + 1}: ${address}\nPrivate key: `) + key
+    chalk.bold(`Address #${i + 1}:  ${address} ${ i == 0 ? firstAccountComment : ''}\nPrivate key: `) + key
```
sohkai: Seems like there's an extra space now at the start of the address?

src/commands/devchain.js
```diff
@@ -99,13 +103,24 @@ exports.printAccounts = (reporter, privateKeys) => {
 }
 
 exports.printMnemonic = (reporter, mnemonic) => {
-  reporter.info(`The accounts were generated from the following mnemonic phrase: ${mnemonic}\n.`)
+  reporter.info(`The accounts were generated from the following mnemonic phrase:\n${mnemonic}\n`)
+}
+
+exports.printResetNotice = (reporter, reset) => {
+  if (reset) {
+    reporter.warning(`The devchain was reset, some steps need to be done to provent issues:
```
sohkai: ~~provent~~ prevent

src/commands/devchain.js
izqui: Vanity feature: addresses and private keys are horizontally aligned to the left now (rather than unaligned by one char).

Issue #151, opened 2018-06-26
It would be nice if there was a way to allow accessing apps from local server


jtremback:
To allow for build tools that watch files and rebuild assets, instead of waiting many seconds for `aragon run` on each change.

jtremback:
Right now it's kind of all related to ipfs, and i'm not sure how to change only the frontend without a full rebuild

jtremback:
How would this be done?

izqui:
@jtremback A flag needs to be added to both `aragon run` and `aragon apm publish` to pass the URL where the app is served (`--http localhost:1234`). If this flag is present, the 'IPFS check' step in `apm publish` can then be disabled. We shouldn't disable it in `aragon run` as other apps hosted in IPFS could be required when running the app with a kit (e.g. interacting with the default voting app that is running from IPFS).

I just did a quick fix to apm.js (https://github.com/aragon/apm.js/pull/22) so that publishing a new version that uses the HTTP provider will work. When publishing, the CLI needs to pass the url of the server as the 'directory'.

An unsolved issue is how is the 'artifact.json' file that is generated before publishing (see https://github.com/aragon/aragon-cli/blob/cf40c307d5179fe3e1f428586c57032c872ef34e/src/commands/apm_cmds/publish.js#L332-L357) is placed in the correct directory so that the file is served on `/artifact.json`. The `manifest.json` file also needs to be served from here.

jtremback:
Hey is this officially working and documented?

Also, @sebohe is taking over on our Aragon integration from me

izqui:
Yes, we just need to add support for it in the CLI

PR #171, opened 2018-08-12
Added HTTP provider feature (WIP)

deamme:
**DO NOT MERGE - STILL WORK IN PROGRESS.**
Fixes https://github.com/aragon/aragon-cli/issues/151

I still have a problem with how to handle `localhost:1234` case because:
https://github.com/aragon/apm.js/blob/master/src/index.js#L35


izqui:
Just created an issue https://github.com/aragon/apm.js/issues/27 for it.

PascalPrecht:
@deamme if you want, you can rebase your work on top of https://github.com/aragon/apm.js/pull/26, that should resolve your issue with `$ apm versions` output.

deamme:
@PascalPrecht Yes, thank you. I made a mistake with this PR unfortunately when merging, so things are not working right now. I'm working on a solution right now

PascalPrecht:
@deamme If you feel like it and want to get rid off the merge commit you can rebase your PR on top of latest `master` instead. Let me know if you need guidance.

deamme:
@PascalPrecht Yeah I should do that, thanks for pointing out. I'm still struggling a bit here though because the wrapper doesn't find any apps even though it's has been published etc. I'm still investigating but I think I need to upgrade the wrapper to use the right apm version too

deamme:
Got things to work  - here's how:
- Aragon.js and its wrapper needs to be updated to reflect new apm. changes
- Paths in manifest file need to change from e.g. `dist/script.js` to `script.js
- Parcel note: One problem with parcel in dev mode is that it apparently stops working if you start the server up first, but starting the parcel in dev mode after the `aragon run...` has finished, works

sohkai:
> Aragon.js and its wrapper needs to be updated to reflect new apm. changes

This is just an update on the `apm.js` dependency, I'm assuming (because of aragon/apm.js#27)?

> Paths in manifest file need to change from e.g. dist/script.js to script.js

This will break the production settings, but maybe we can force this as a override when loading apps via http.

izqui:
> This is just an update on the apm.js dependency

@sohkai yes, apm.js v2.0.1 was released last week: https://github.com/aragon/apm.js/releases/tag/v2.0.1 Since the dependency version is not pinned to 2.0.0 on neither aragon.js nor the wrapper, it should just work when installing now.

> This will break the production settings, but maybe we can force this as a override when loading apps via http.

I don't follow why that would be the case.

sohkai:
@ji I was unclear by "production" settings... I meant the way the app is published to IPFS.

As it seems we're expecting to load files from `<ipfs hash>/dist/script.js` (if that's what's in the `manifest.json`), changing it to `<ipfs hash>/script.js` will break unless the publishing scripts accomodates that changes too.

deamme:
@sohkai Should be alright now

izqui:
@deamme I have been testing this locally and it works as expected, great contribution! Made some minor changes to your branch and updated the apm.js dependency.

src/commands/run.js
```diff
@@ -72,8 +72,8 @@ exports.builder = function (yargs) {
   }).option('http', {
     description: 'URL for where your app is served from e.g. localhost:1234',
     default: null,
-  }).option('served-at', {
-    description: 'URL for where your app is served from e.g. localhost:1234',
+  }).option('http-served-from', {
```
izqui: Changed the flag `--served-at` to `--http-served-from` so it is more descriptive to what the flag does.

src/commands/run.js
```diff
@@ -69,6 +69,12 @@ exports.builder = function (yargs) {
   }).option('build-script', {
     description: 'The npm script that will be run when building the app',
     default: 'build',
+  }).option('http', {
+    description: 'URL for where your app is served from e.g. localhost:1234',
+    default: null,
+  }).option('http-served-from', {
```
izqui: Changed the flag `--served-at` to `--http-served-from` so it is more descriptive to what the flag does.

PR #174, opened 2018-08-16
uiux(commands/apm): show meaningful error messages

### LINKED ISSUE #167: `apm versions` shouldn't fail if one of the repo version's contentURI is invalid
Currently if one of the published versions in a repo had a badly formatted content URI, doing `aragon apm versions` for the repo results in:

```
✖ Invalid content URI (expected format was "<provider>:<identifier>")
```

The CLI should show the versions that have a correct format as usual, and print a warning sign and the contentURI string for the ones that have an invalid format.

This will require making changes in https://github.com/aragon/apm.js as well, so it doesn't just throw an error when one of the version's content URI is bad.

PascalPrecht:
As discussed in #167, `apm versions` throws when any of the returned
content URIs are invalid. This behaviour will be fixed with https://github.com/aragon/apm.js/pull/26,
so now we're making use of that exposed error or warning message for
a better user experience.

Closes #167

### DIFF
--- a/src/commands/apm_cmds/versions.js
+++ b/src/commands/apm_cmds/versions.js
@@ -10,12 +10,13 @@ exports.handler = async function ({ reporter, module, bump, cwd, network, apm: a
   const web3 = await ensureWeb3(network)
 
   apmOptions.ensRegistryAddress = apmOptions['ens-registry']
-
   const versions = await APM(web3, apmOptions).getAllVersions(module.appName)
   reporter.info(`${module.appName} has ${versions.length} published versions`)
   versions.map((version) => {
     if (version && version.content) {
       reporter.success(`${version.version}: ${version.contractAddress} ${version.content.provider}:${version.content.location}`)
+    } else if (version && version.error) {
+      reporter.warning(`${version.version}: ${version.contractAddress} ${version.error}`)
     } else {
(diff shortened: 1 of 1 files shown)


src/commands/apm_cmds/versions.js
```diff
@@ -10,12 +10,13 @@ exports.handler = async function ({ reporter, module, bump, cwd, network, apm: a
...
 
   apmOptions.ensRegistryAddress = apmOptions['ens-registry']
-
   const versions = await APM(web3, apmOptions).getAllVersions(module.appName)
   reporter.info(`${module.appName} has ${versions.length} published versions`)
   versions.map((version) => {
     if (version && version.content) {
       reporter.success(`${version.version}: ${version.contractAddress} ${version.content.provider}:${version.content.location}`)
+    } else if (version && version.error) {
+      reporter[version.reportType](`${version.version}: ${version.contractAddress} ${version.error}`)
```
PascalPrecht: `reportType` is exposed in https://github.com/aragon/apm.js/pull/26 respectively.

izqui: As commented in the apm.js issue, we can just report a warning in both cases, and break the current coupling of apm.js having to be aware of the supported reporters in the aragon-cli

PR #175, opened 2018-08-16
fix(apm/publish): check if artifact.json already exist

### LINKED ISSUE #159: apm publish should sanity check if an artifact.json and manifest.json exist in the publishing directory
Ideally it would warn the user and ask if they want to continue.

PascalPrecht:
@sohkai just a first stab on solving #159 and figuring out whether this is the direction you were aiming for. I'm not super familiar with the publishing process yet and still have to see what's the best way to test this locally, but hopefully this gives us a base for discussing the desired behavior.

See my inline comments.

Closes #159


src/commands/apm_cmds/publish.js
```diff
@@ -349,10 +350,15 @@ exports.task = function ({
       title: 'Generate application artifact',
       skip: () => onlyContent && !module.path, // TODO: If onlyContent has been set, get previous version's artifact
       task: async (ctx, task) => {
-        const dir = onlyArtifacts ? cwd : ctx.pathToPublish
-        const artifact = await generateApplicationArtifact(web3, cwd, dir, module, contract, reporter)
-
-        return `Saved artifact in ${dir}/artifact.json`
+        if (pathExistsSync(`${ctx.pathToPublish}/${MANIFEST_FILE}`) &&
+            pathExistsSync(`${ctx.pathToPublish}/${ARTIFACT_FILE}`)) {
+          task.skip('Reusing existing artifact');
```
PascalPrecht: I first thought we might wanna take advantage of the dedicated `skip()` function here, but it seems that API doesn't give us access to `ctx` which we need to look at the correct path. That's my I went for `task.skip()` within the `task()` handler.

Also notice that there's no warning for the user with an option yet to decide whether an existing artifact should be reused or not. I assume this can be easily added with modules like [prompt-choice](https://www.npmjs.com/package/prompt-choices). I first wanted to verify and double check with you if this is the right places for those changes.

Sidenote: I'm unaware of what `onlyContent` in the context of this command means, if you have some insights on that, that'd be great for me for better understanding :)

src/commands/apm_cmds/publish.js
```diff
@@ -349,10 +350,15 @@ exports.task = function ({
...
+        } else {
```
PascalPrecht: We might not need this `else` block `task.skip()` prevents further execution

sohkai:
@PascalPrecht Thanks for working on this! I think for the functionality currently in this PR (skipping artifact generation if one already exists), it'd be a great use of a user prompt.

However, #159 is actually about getting `aragon apm publish --only-artifact` to detect that the current directory doesn't have a artifact.json when someone tries to deploy, and show the user a warning prompt (to either abort or generate it).

This was an issue when I tried to deploy some of our apps, without realizing I was missing the artifact.json (see section for `vault.aragonpm.eth` in https://github.com/aragon/deployments/blob/master/rinkeby/README.md#aragonpmeth-repo-updates-1)

PascalPrecht:
@sohkai thanks for getting back to me here. I'm back from vacation now and will look into this.

sohkai:
@PascalPrecht Sorry, I definitely got things mixed up in that last reply 😅.

I think what I meant to say was the `--only-content` flag, like in this publish command:

```sh
apm publish 0x0000000000000000000000000000000000000000 --network rinkeby --apm.ens-registry "0xfbae32d1cde62858bc45f51efc8cc4fa1415447e" --no-ipfs-check --apm.ipfs.rpc "http://ipfs.aragon.network:5001" --only-content --no-build
```

In the [linked deployment notes for the Vault](https://github.com/aragon/deployments/blob/master/rinkeby/README.md#aragonpmeth-repo-updates-1), I had to first build the artifact.json via `--only-artifacts` and then publish via `--only-content`.

When I initially published `1.0.4` with `--only-content`, I hadn't generated the artifact.json yet, and so the deployment was incomplete.

PascalPrecht:
@sohkai and all of a sudden, things make sense! Will update the PR accordingly. 

Unfortunately, `Listr` eats anything that's being send to stdout within tasks, making it kinda impossible to implement a prompt that asks the user for action. Will have to dig deeper on this one to figure out why that is and if and how we can work around that...

PascalPrecht:
Okay, so Listr supports [renderers](https://github.com/SamVerschueren/listr#custom-renderers) that control how output is being rendered. It also comes with a couple of pre-implemented ones, one of them being the [verbose-renderer](https://github.com/SamVerschueren/listr-verbose-renderer). 

As you might've guessed, it pretty much renders everything that's being emitted by tasks and it also works correctly with an integrated prompt.

The downside however is that really prints out everything, so I think, if we'd want the nice looking task list + a nice prompt, but no verbose output, we'll have to build our own renderer. Unfortunately, there's no way to distinguish between "normal" stdout data and a prompt, so I'm not sure if that can be solved with the available APIs without going a down a super hacky road.

There's two other options I could think of to get this feature in and move on (for now):

1. We don't do a prompt at all and simply abort with an error message if no artifact exists, also telling the user to generate one first (and how).
2. We could perform the check before the tasklist is even initialized, have a prompt and either exit early (before the task list is initialized), or always generate the artifact respectively.

PascalPrecht:
Just realized that option 2) requires `dirToPublish` to exist, which in the case of using `--only-content` is a temporary folder generated as part of the task. Why might not want to do this work up front, so I guess option 1) would be the most pragmatic to move forward with this one for now.

PascalPrecht:
Okay, I now went ahead and implemented 2), simply throwing with a meaningful error message in case `artifact.json` doesn't exist.

I've also remove the first implementations to let the user know that an existing artifact is being reused. The reason being is that

1. I misunderstood the original issue (basically I was solving a different problem that isn't really a problem)
2. Even if we'd keep it, we'd want a prompt there as well to ask the user whether the artifact should be re-generated or not. This is something that doesn't seem to be feasible to implement right now, as mentioned in https://github.com/aragon/aragon-cli/pull/175#issuecomment-418674574

If you do prefer getting that stuff in, let me know and we can revisit.

PascalPrecht:
I did a bit [more research](https://github.com/SamVerschueren/listr/issues/105) and it turned out that there is already a package that adds rudimentary prompt input support to listr. I've added another commit that adds it to the `publish` command. Here's what it looks like:

sep-10-2018 11-20-46

I think this should do it. Let me know if this needs more chance (probably changes in wording are required. 

Also, I'll remove the first commit in case we decide going for the version with the prompt.

src/commands/apm_cmds/publish.js
```diff
@@ -392,9 +398,22 @@ exports.task = function ({
       skip: () => onlyContent && !module.path, // TODO: If onlyContent has been set, get previous version's artifact
       task: async (ctx, task) => {
         const dir = onlyArtifacts ? cwd : ctx.pathToPublish
-        const artifact = await generateApplicationArtifact(web3, cwd, dir, module, contract, reporter)
 
-        return `Saved artifact in ${dir}/artifact.json`
+        if (onlyContent && !pathExistsSync(`${dir}/${ARTIFACT_FILE}`)) {
```
izqui: In a publishing without `--only-content`, which means the contract is being updated, we should always generate a new artifact even if it exists already.

With the current changes, it would never generate artifacts unless the `--only-content` flag is provided.

PascalPrecht: 👍

PascalPrecht:
@izqui updated the PR as requested.

sohkai:
@PascalPrecht Looks really good! Thanks for doing all that research, and I think it's paved the way for getting a few more annoying things with the CLI out of the way :).

As for

> I misunderstood the original issue (basically I was solving a different problem that isn't really a problem)

I'm not sure if this is useful, as AFAIK the artifacts are generally meant to be overridden, but it might be nice as a user option. It should be done separately in another PR though :).

PascalPrecht:
>  It should be done separately in another PR though :).

Agreed.

I've rebased this one on latest master. This should be ready for merge unless there's any more requests for change.

PR #176, opened 2018-08-16
fix(commands/apm/grant): use updated CREATE_VERSION_ROLE hash

### LINKED ISSUE #158: Update APMRepo create version role hash
When doing `aragon apm grant 0xab` for giving the `CREATE_VERSION_ROLE` to `0xab`, it will do it for the role hash `0x00...01` which was used in the previous versions of aragonOS.

The current correct role hash is `keccak256('CREATE_VERSION_ROLE')`.

For this not to be a backwards incompatible change (there are instances of APM with this role hash deployed on rinkeby), we should add a flag that allows for setting the permission with the legacy role hash(`--legacy-role`). Or even better, a call can be made to the Repo contract to get the id for the role.

PascalPrecht:
As discussed in #158, `apm grant 0xsomeAddress` uses a create version
role id hash that is no longer be used by Repo contracts. This commit
ensures that the new role id is used by fetching it from the dedicated
repo address.

Also, this might be considered a breaking change as the new default for the role id, is `keccak256('CREATE_VERSION_ROLE')` and no longer `0x....01`

Fixes #158

### DIFF
--- a/src/acl/index.js
+++ b/src/acl/index.js
@@ -8,11 +8,16 @@ module.exports = (web3) => {
     return new web3.eth.Contract(require('@aragon/os/build/contracts/ACL').abi, aclAddr)
   }
 
+  const getRoleId = async (repoAddr) => {
+    const repo = new web3.eth.Contract(require('@aragon/os/build/contracts/Repo').abi, repoAddr)
+    return await repo.methods.CREATE_VERSION_ROLE().call()
+  }
+
   return {
     grant: async (repoAddr, grantee) => {
       const acl = await getACL(repoAddr)
 
-      const roleId = '0x0000000000000000000000000000000000000000000000000000000000000001'
+      const roleId = await getRoleId(repoAddr)
 
       const call = acl.methods.grantPermission(grantee, repoAddr, roleId)
       return {
--- a/src/commands/apm_cmds/grant.js
+++ b/src/commands/apm_cmds/grant.js
@@ -20,10 +20,10 @@ exports.handler = async function ({
   network,
   module,
   apm: apmOptions,
-
   // Arguments
   address
 }) {
+
   const web3 = await ensureWeb3(network)
   apmOptions.ensRegistryAddress = apmOptions['ens-registry']
 
--- a/src/commands/apm_cmds/publish.js
+++ b/src/commands/apm_cmds/publish.js
@@ -29,7 +29,7 @@ exports.describe = 'Publish a new version of the application'
 exports.builder = function (yargs) {
   return deploy.builder(yargs) // inherit deploy options
     .positional('contract', {
-      description: 'The address or name of the contract to publish in this version. If it isn\' provided, it will default to the current version\'s contract.'
+      description: 'The address or name of the contract to publish in this version. If it isn\'t provided, it will default to the current version\'s contract.'
     }).option('only-artifacts', {
       description: 'Whether just generate artifacts file without publishing',
       default: false,


src/acl/index.js
```diff
@@ -8,11 +8,18 @@ module.exports = (web3) => {
...
+
   return {
-    grant: async (repoAddr, grantee) => {
+    grant: async (repoAddr, grantee, useLegacyRoleId = false) => {
       const acl = await getACL(repoAddr)
 
-      const roleId = '0x0000000000000000000000000000000000000000000000000000000000000001'
+      const roleId = useLegacyRoleId ?
+        '0x0000000000000000000000000000000000000000000000000000000000000001' :
+        await getRoleId(repoAddr)
```
PascalPrecht: @izqui in #158 you've said:

> Or even better, a call can be made to the Repo contract to get the id for the role.

While this sounds more like an either/or scenario, I've now implemented both ideas: getting hash id from the contract and explicitly falling back to the legacy id using `--legacy-role`.

It wasn't entirely clear to me if both were desired or just one of those options. Let me know if you want this to be changed.

src/acl/index.js
```diff
@@ -8,11 +8,18 @@ module.exports = (web3) => {
     return new web3.eth.Contract(require('@aragon/os/build/contracts/ACL').abi, aclAddr)
   }
 
+  const getRoleId = async (repoAddr) => {
+    const repo = new web3.eth.Contract(require('@aragon/os/build/contracts/AragonApp').abi, repoAddr)
```
izqui: It should be the `Repo` contract rather than `AragonApp`

src/cli.js
```diff
@@ -92,7 +92,11 @@ cmd.option('apm.ens-registry', {
   description: 'Address of the ENS registry',
   default: require('@aragon/aragen').ens
 })
-cmd.group(['apm.ens-registry', 'eth-rpc'], 'APM:')
+cmd.option('legacy-role', {
```
izqui: As commented above it is better to just remove the option and always check the role hash in the contract. However this option would have been better located in the command options directly, as it doesn't need to be a global CLI setting.

https://github.com/aragon/aragon-cli/blob/a7822caef299ea120768d5526b495635ba304b0a/src/commands/apm_cmds/grant.js#L10-L14

src/acl/index.js
izqui: If we get the role hash directly from the contract there is no need to have an option for supporting the legacy role hash, as the contract will always return the actual role hash for the repo. Sorry the issue description wasn't clear enough 😅

I think we should drop the legacy flag and always use the role of the contract.

src/acl/index.js
```diff
@@ -8,11 +8,18 @@ module.exports = (web3) => {
...
+    const roleId = await repo.methods.CREATE_VERSION_ROLE().call()
```
izqui: Missing a return here

src/acl/index.js
PascalPrecht: Sure! That's what happens when you don't test this on your local machine :) 

Will fix this.

src/acl/index.js
PascalPrecht: Makes sense. I'll remove it.

src/acl/index.js
PascalPrecht: Good catch!

izqui:
The way I have tested it is:

```
aragon init test-app
aragon run
apm grant [granted address]
```

After `aragon run` either keep that running or stop it and do `aragon devchain` so the grant command can connect to the running chain.

PascalPrecht:
Thanks for the review @izqui !

Most of your comments boil down to lack of local testing. I'll dig into how to setup a local environment where I can run the code against.

izqui:
@PascalPrecht ping me in aragon.chat if you need help setting it up!

PascalPrecht:
Thank you!!

src/acl/index.js
PascalPrecht: I'm definitely missing what the granting permission in the context of this command and the system means and how all the smart contracts are related to each other, however, maybe you can give me a little pointer on why `repoAddr` is used on an `AragonApp` contract here: https://github.com/aragon/aragon-cli/blob/master/src/acl/index.js#L3 (instead of `Repo` contract)

izqui: There we just need to use the `kernel()` method which `Repo` inherits from `AragonApp`: https://github.com/aragon/aragonOS/blob/dev/contracts/apm/Repo.sol#L6. That could also be `Repo` instead of `AragonApp` and would still work.

izqui:
Tested on both rinkeby and local network and it is working as expected!

PR #194, opened 2018-09-10
feat(commands/run): introduce --client-version and --client-port 

PascalPrecht:
The important commit here is the third one. It might make sense to squash them all into a single one, in case you want to preserve the refactoring in the commit history.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,5 +90,8 @@
     "require": [
       "@babel/register"
     ]
+  },
+  "aragon": {
+    "clientVersion": "fcebf55d1af6c574a92587e2a1d3ac8c00804d16"
   }
 }
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     ]
   },
   "aragon": {
-    "clientVersion": "fcebf55d1af6c574a92587e2a1d3ac8c00804d16"
+    "clientVersion": "fcebf55d1af6c574a92587e2a1d3ac8c00804d16",
+    "clientPort": "3000"
   }
 }
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -18,6 +18,7 @@ const os = require('os')
 const fs = require('fs-extra')
 const opn = require('opn')
 const execa = require('execa')
+const pkg = require('../../package.json')
 
 const {
   findProjectRoot,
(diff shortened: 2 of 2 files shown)


src/commands/run.js
```diff
@@ -231,26 +236,25 @@ exports.handler = function ({
...
             // Clone wrapper
-            return clone(
-              'https://github.com/aragon/aragon',
-              WRAPPER_PATH,
-              { checkout: WRAPPER_COMMIT }
-            )
-          },
+            return clone('https://github.com/aragon/aragon', CLIENT_PATH, {
+              checkout: clientVersion
+            })
```
PascalPrecht: If the checkout fails because an invalid version is passed down to this command, CLI will fail with

```
'git checkout' failed with status code 1
```

I thought it'd be nicer to throw a more descriptive error message, letting the user know that the given version probably doesn't exist, however, hooking into `clone()`'s callback to rethrow the error, results in TaskList actually finishing its tasks successfully, with the error being throw slightly later (as it's async).

This is very likely related to how Listr expects errors to be thrown but needs more investigation.

src/commands/run.js
```diff
@@ -85,6 +86,9 @@ exports.builder = function (yargs) {
     description: 'Arguments for calling the app init function',
     array: true,
     default: [],
+  }).option('client-version', {
```
PascalPrecht: `--client-app-version` ?

src/commands/run.js
```diff
@@ -32,9 +33,9 @@ const { Writable } = require('stream')
 const url = require('url')
 
 const TX_MIN_GAS = 10e6
-const WRAPPER_PORT = 3000
+const CLIENT_PORT = 3000
```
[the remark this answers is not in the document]
PascalPrecht: As discussed offline, it turns out `CLIENT_PORT` is not actually used when the aragon client is started.

More info: https://github.com/aragon/aragon-cli/issues/198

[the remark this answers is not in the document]
PascalPrecht: I propose fixing that in a separate PR so that this one isn't dependent on coordinating updates with CLI and aragon/aragon.

PascalPrecht:
@izqui I've updated this PR to use the default client version as default value.

I also went ahead and moved the client port out of the code base to make it configurable via `package.json` and a newly introduced `--client-port` option.

Notice however that, as mentioned in https://github.com/aragon/aragon-cli/pull/194#discussion_r216675380, this options is being ignored (and has always been).

I'll address that in an additional PR.

package.json
```diff
@@ -90,5 +90,9 @@
     "require": [
       "@babel/register"
     ]
+  },
+  "aragon": {
+    "clientVersion": "fa2025232330bc3e4d3b792cab4bf44d754d33e6",
+    "clientPort": "7000"
```
izqui: Just a reminder so we don't forget reverting the default port back to 3000

PascalPrecht:
@izqui updated.

PR #197, opened 2018-09-11
Implement version bump option for apm publish

PascalPrecht:
Please find detailed descriptions in the dedicated commits of this PR.

### DIFF
--- a/src/commands/apm_cmds/publish.js
+++ b/src/commands/apm_cmds/publish.js
@@ -18,11 +18,13 @@ const execa = require('execa')
 const { compileContracts } = require('../../helpers/truffle-runner')
 const web3Utils = require('web3-utils')
 const deploy = require('../deploy')
+const versionTask = require('./version')
 const startIPFS = require('../ipfs')
 const getRepoTask = require('../dao_cmds/utils/getRepoTask')
 
 const MANIFEST_FILE = 'manifest.json'
 const ARTIFACT_FILE = 'artifact.json'
+const VERSION_BUMP_TYPES = ['patch', 'minor', 'major']
 
 exports.command = 'publish [contract]'
 
@@ -76,6 +78,9 @@ exports.builder = function (yargs) {
     }).option('http-served-from', {
       description: 'Directory where your files is being served from e.g. ./dist',
       default: null,
+    }).option('version-bump', {
+      description: 'Specify version bump type for to be published app',
+      default: null
     })
 }
 
@@ -192,7 +197,7 @@ exports.task = function ({
   key,
   files,
   ignore,
-  automaticallyBump,
+  versionBump,
   ipfsCheck,
   publishDir,
   init,
@@ -203,19 +208,36 @@ exports.task = function ({
   http,
   httpServedFrom,
 }) {
-  if (onlyContent) {
-    contract = '0x0000000000000000000000000000000000000000'
-  }
   
   apmOptions.ensRegistryAddress = apmOptions['ens-registry']
   const apm = APM(web3, apmOptions)
   return new TaskList([
     {
       title: 'Preflight checks for publishing to APM',
-      enabled: () => !automaticallyBump,
       task: (ctx) => new TaskList([
         {
+          title: 'Fetch latest app version from chain and increment',
+          enabled: () => !!versionBump,
+          task: async (ctx, task) => {
+
+            if (VERSION_BUMP_TYPES.indexOf(versionBump) == -1) {
+              throw new Error(`Not a valid version bump type. Please use one of 'patch', 'minor', 'major'`)
+            }
+
+            const newVersion = await versionTask.bumpVersionAndUpdateManifest({
+              module,
+              bump: versionBump,
+              cwd,
+              apm
+            })
+
+            ctx.version = newVersion
+            ctx.isMajor = versionBump === VERSION_BUMP_TYPES[2]
+          }
+        },
+        {
           title: 'Check version is valid',
+          enabled: () => !!!versionBump,
           task: (ctx) => {
             if (module && semver.valid(module.version)) {
               ctx.version = module.version
@@ -230,6 +252,7 @@ exports.task = function ({
         },
         {
           title: 'Checking version bump',
+          enabled: () => !!!versionBump,
           task: async (ctx) => {
             let repo = { version: '0.0.0' }
             try {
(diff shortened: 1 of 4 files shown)


src/commands/apm_cmds/publish.js
```diff
@@ -208,8 +211,30 @@ exports.task = function ({
...
+        {
+          title: 'Fetch latest app version from chain and increment',
+          enabled: () => !!versionBump,
+          task: async (ctx, task) => {
+
+            if (VERSION_BUMP_TYPES.indexOf(versionBump) == -1) {
+              throw new Error(`Not a valid version bump type. Please use one of 'patch', 'minor', 'major'`)
+            }
+
+            const newVersion = await versionTask.bumpVersionAndUpdateManifest({
```
PascalPrecht: This API now always resolves with a version either way. Please review this carefully.

src/commands/apm_cmds/version.js
```diff
@@ -1,7 +1,31 @@
-const fs = require('fs')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const fs = require('fs-extra')
```
PascalPrecht: Switched to fs-extra as it's basically fs with flowers. Plus it has been around already anyways.

src/commands/apm_cmds/version.js
```diff
@@ -1,7 +1,31 @@
...
+const bumpVersionAndUpdateManifest = async ({ module, bump, network, apmOptions, cwd }) => {
+
+  const web3 = await ensureWeb3(network)
+  const apm = APM(web3, apmOptions)
+
+  let repo = { version: '0.0.0' }
+  try {
+    repo = await apm.getLatestVersion(module.appName)
+  } catch (e) {
+
```
PascalPrecht: Notice that we swallow possible errors here. So far the only case I could imagine was that `getLatestVersion()` fails due to the very first deployment. This however shouldn't propagate an error because we know we can then just go ahead with an initial 1.0.0 version.

Let me know if this needs adjustment, e.g. cases where this API fails without being the very first publish.

src/commands/apm_cmds/version.js
```diff
@@ -1,7 +1,31 @@
...
+  }
+
+  const manifestLocation = await findUp('arapp.json', { cwd })
+  const manifest = fs.readJsonSync(manifestLocation)
+
+  manifest.version = semver.valid(bump) ? semver.valid(bump) : semver.inc(repo.version, bump)
+
+  fs.writeJsonSync(manifestLocation, manifest)
```
izqui: We should still indent the JSON file as before

src/commands/apm_cmds/publish.js
```diff
@@ -74,6 +76,9 @@ exports.builder = function (yargs) {
     }).option('http-served-from', {
       description: 'Directory where your files is being served from e.g. ./dist',
       default: null,
+    }).option('version-bump', {
+      description: 'Specify version bump type for to be published app',
+      default: null
```
izqui: We can use yargs [choices](http://yargs.js.org/docs/#api-choiceskey-choices) so only values in the VERSION_BUMP_TYPES array are valid.

src/commands/apm_cmds/publish.js
```diff
@@ -251,24 +277,35 @@ exports.task = function ({
             }
```
izqui: We can edit the error messages above pointing to using the new `--version-bump` flag

src/commands/apm_cmds/publish.js
```diff
@@ -251,24 +277,35 @@ exports.task = function ({
...
+        if (!ctx.isMajor || onlyContent) {
+          return 'Publishing patch or minor version, skipping contract compilation and deployment.'
+        }
       },
-      enabled: ctx => !onlyContent && ((contract && !web3Utils.isAddress(contract)) || (!contract && ctx.isMajor && !reuse) || automaticallyBump)
+      task: ctx => new TaskList([
+        {
+          title: 'Compile contracts',
+          task: async () => (await compileContracts()),
+          enabled: () => !onlyContent && web3Utils.isAddress(contract)
```
izqui: Can be simplified to just `web3Utils.isAddress(contract)` as we already skip if running in `onlyContent` mode

src/commands/apm_cmds/publish.js
izqui: I actually just tested this and the error message is quite crappy and would also require us to add `null` to the `VERSION_BUMP_TYPES` array. We can leave it as it is
2018-09-11 at 4 32 25 pm

src/commands/apm_cmds/publish.js
```diff
@@ -251,24 +277,35 @@ exports.task = function ({
...
+        },
+        {
+          title: 'Deploy contract',
+          task: async (ctx) => {
+            const deployTaskParams = { contract, init, reporter, network, cwd, web3, apmOptions }
+
+            return await deploy.task(deployTaskParams)
+          },
+          enabled: ctx => !onlyContent && ((contract && !web3Utils.isAddress(contract)) || (!contract && ctx.isMajor && !reuse) || automaticallyBump)
```
izqui: Can be simplified to `!web3Utils.isAddress(contract) && !reuse` 🎉

src/commands/apm_cmds/publish.js
```diff
@@ -251,24 +277,35 @@ exports.task = function ({
...
+        },
+      ])
     },
     {
       title: 'Automatically bump version',
```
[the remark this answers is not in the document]
PascalPrecht: Not sure I follow. You want the automatically bump task to be removed in favor of setting `versionBump`'s default to `true`?

src/commands/apm_cmds/publish.js
PascalPrecht: The thing with this error message is that it's getting pretty long and is already actually not readable when it's being rendered by the CLI (as it gets cut off). 

I'll see what I can do.

src/commands/apm_cmds/publish.js
PascalPrecht: 👍

src/commands/apm_cmds/publish.js
PascalPrecht: 👍

src/commands/apm_cmds/version.js
```diff
@@ -1,7 +1,31 @@
-const fs = require('fs')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const fs = require('fs-extra')
 const findUp = require('find-up')
 const semver = require('semver')
+const APM = require('@aragon/apm')
 
+const bumpVersionAndUpdateManifest = async ({ module, bump, network, apmOptions, cwd }) => {
+
+  const web3 = await ensureWeb3(network)
```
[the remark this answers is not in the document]
PascalPrecht: Yes, I considered that too. I wasn't sure if that'd be a nice DX though because it would require consumers to always create an APM instance manually first.

On the other hand, since this function is most likely be used within other commands, an APM instance prolly always available.

src/commands/apm_cmds/version.js
```diff
@@ -1,7 +1,31 @@
...
+    repo = await apm.getLatestVersion(module.appName)
+  } catch (e) {
+
+  }
+
+  const manifestLocation = await findUp('arapp.json', { cwd })
+  const manifest = fs.readJsonSync(manifestLocation)
+
+  manifest.version = semver.valid(bump) ? semver.valid(bump) : semver.inc(repo.version, bump)
+
```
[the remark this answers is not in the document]
PascalPrecht: @izqui can you elaborate in which scenario `manifest.version` would be null?

src/commands/apm_cmds/version.js
PascalPrecht: Yes, good point! 👍

src/commands/apm_cmds/version.js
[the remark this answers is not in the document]
izqui: If bump is not a version number nor a valid bump (`major`, `minor` or `patch`).

src/commands/apm_cmds/publish.js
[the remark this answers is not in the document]
izqui: `automaticallyBump` should no longer be needed if the run command always sets `versionBump` to `major`

src/commands/apm_cmds/version.js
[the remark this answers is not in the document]
izqui: When used inside the `version` command I would expect that `apm` and `web3` get initialized in the [handler](https://github.com/aragon/aragon-cli/pull/197/files/99254b4c0d321ef198a0c69d93897a34badf9111#diff-0aaf30104993c9a323278a45ec79650dR42).

src/commands/apm_cmds/publish.js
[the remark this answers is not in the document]
PascalPrecht: Got it.

src/commands/apm_cmds/version.js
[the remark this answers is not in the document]
PascalPrecht: 👍

src/commands/apm_cmds/version.js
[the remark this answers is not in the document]
PascalPrecht: Got it

PascalPrecht:
Updated the commits and rebased them accordingly. This should address all of your comments.

PTAL

src/commands/apm_cmds/publish.js
```diff
@@ -74,6 +76,9 @@ exports.builder = function (yargs) {
     }).option('http-served-from', {
       description: 'Directory where your files is being served from e.g. ./dist',
       default: null,
+    }).option('version-bump', {
+      description: 'Specify version bump type for to be published app',
+      default: 'major'
```
izqui: I would introduce the `--version-bump` flag without a default value for now, requiring bumping on publish to be explicit.

We can still make that when `aragon run` uses the publish command as part of its execution, it always does a major bump as the old `automaticallyBump` did.

izqui: https://github.com/aragon/aragon-cli/blob/6137e8f4987ddfcf7f85553641b3dd927badf367/src/commands/run.js#L168

PascalPrecht:
Updated :)

izqui:
Closing in favor of #230.

PR #200, opened 2018-09-11
feat(commands/run): introduce --client-path option

### LINKED ISSUE #199: Introduce option use custom locally installed Aragon client
This came up as I was implementing #198. In order to implement #198, we need to make changes in aragon/aragon's npm scripts as it should look for a newly introduced `ARAGON_CLIENT_PORT` env variable.

While setting custom env variables through the CLI works, there's no easy way to verify whether the installed Aragon client (wrapper) actually consumes that env variable. 

To verify this, we need a way to make CLI use a custom, locally installed version of Aragon client that hasn't been published yet (similar to what we do with npm link). After discussing this with @izqui, we decided to introduce a new option (maybe `---client-path`) that lets users point the CLI to an already locally installed version of the client:

```
$ aragon run --client-path=~/project/some/path --client-port=4000
```

PascalPrecht:
This commit enables users to pass a `--client-path` to the `run` command.
The path specifies the location to an already existing installation of the
Aragon client on the local machine, allowing developers to test custom versions
of the app.

Closes #199

### DIFF
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -92,6 +92,9 @@ exports.builder = function (yargs) {
   }).option('client-port', {
     description: 'Port being used by Aragon client',
     default: DEFAULT_CLIENT_PORT
+  }).option('client-path', {
+    description: 'A path pointing to an existing Aragon client installation',
+    default: null
   })
 }
 
@@ -131,7 +134,8 @@ exports.handler = function ({
     appInit,
     appInitArgs,
     clientVersion,
-    clientPort
+    clientPort,
+    clientPath
   }) {
 
   apmOptions.ensRegistryAddress = apmOptions['ens-registry']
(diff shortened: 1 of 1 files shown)


izqui:
@PascalPrecht Needs rebasing after merging #194

PascalPrecht:
@izqui done

src/commands/run.js
```diff
@@ -132,6 +135,7 @@ exports.handler = function ({
     appInitArgs,
     clientVersion,
     clientPort
```
izqui: Missing a `,` here

PascalPrecht: Good catch! 👍 

Slipped through during rebase.

PascalPrecht: This is now fixed.

Issue #203, opened 2018-09-12
Improving developer environment setup


luisivan:
With a 30MB/s connection and a i7 laptop, it takes:
- 1 minute to `npm i -g aragon-cli`
- 1 minute and 30 seconds to `aragon init`
- 1 minute to `aragon run` until downloading the client
- 1 minute and 30 seconds to download the client and install its dependencies
- 30 seconds to start the client

Which adds up to **5 minutes and 30 seconds**. Plus whatever it takes to install IPFS.
I'm guessing lots of devs will drop out in the process. This could also be worse in places in the world where they cannot afford fast computers and lots of bandwidth.

Thought of some suggestions on how to improve it.
#### Expected time gain: 1 minute
Devs are installing `aragon-cli`, only to install it again when they `aragon run`. If a dev wants to pin their package to a specific `aragon-cli` version, they could add it manually. But I think as default it creates a lot of overhead and also is inconsistent, since I'd expect to use the same version of the tool if they are asking me to install it globally.
_PRs: https://github.com/aragon/aragon-bare-boilerplate/pull/10 https://github.com/aragon/aragon-react-boilerplate/pull/26 https://github.com/aragon/aragon-react-kit-boilerplate/pull/14_

#### Expected time gain: 30 seconds to 1 minute

Spin off `aragon init` into `create-aragon-app`, so users can just use `npx create-aragon-app <name> <boilerplate>` rather than having to install `@aragon/cli` globally.

#### Expected time gain: 2 minutes
The client has a lot of dependencies, but if you are not using your own one, it doesn't add anything to have to download them all. If you are using the default one, it doesn't even make sense that you have to compile it locally. We could offer the final build of the client, which would remove the need of installing dependencies and compiling the client.

#### Expected time gain: 30 seconds
Installing `go-ipfs` is not very easy because it's not in npm, therefore people need to jump to a website and figure out how to install it. We could use `js-ipfs`, which took a minute to install in my machine.
_Deprecated in favor of pre-bundling `go-ipfs`_

#### Expected time gain: 2 minutes
We could include the binaries in a build, and download them when installing `aragon-cli`. If we detect they are already installed, we could skip this step.
_PR: https://github.com/aragon/aragon-cli/pull/205_

#### Expected time gain: 1 minute
We should pre-package the known roles when building the CLI, so users don't have to install all aragon-apps.
_PR: https://github.com/aragon/aragon-cli/pull/209_

#### Expected time gain: 1 minute
[...]

PR #205, opened 2018-09-12
Install go-ipfs as dependency, remove the need for manual install

luisivan:
Addresses the 4th point at https://github.com/aragon/aragon-cli/issues/203, and removes the need for manually installing IPFS.

### DIFF
--- a/src/commands/ipfs.js
+++ b/src/commands/ipfs.js
@@ -30,7 +30,7 @@ exports.task = ({ apmOptions }) => {
 					} else {
 						task.output = 'IPFS is started, checking CORS config'
 						await setIPFSCORS(apmOptions.ipfs.rpc)
-						return 'Connected to IPFS daemon ar port: '+ apmOptions.ipfs.rpc.port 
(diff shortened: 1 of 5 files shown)


src/commands/ipfs.js
```diff
@@ -22,25 +21,17 @@ exports.task = ({ apmOptions }) => {
...
+					if (!running) {
+						task.output = 'Starting IPFS at port: ' + apmOptions.ipfs.rpc.port
+						await startIPFSDaemon()
+						ctx.started = true
+						await setIPFSCORS(apmOptions.ipfs.rpc)
+					} else {
+						task.output = 'IPFS is started, checking CORS config'
+						await setIPFSCORS(apmOptions.ipfs.rpc)
+						return 'Connected to IPFS daemon ar port: '+ apmOptions.ipfs.rpc.port 
+					}
```
PascalPrecht: I think this is little too far indented no?

luisivan: Indeed, my code editor replaced spaces with tabs. Fixed it now

PR #210, opened 2018-09-18
Update @aragon/apm to 2.0.2

izqui:
Fixes gas limit not being an integer, also decided on passing down the error from apm.js directly rather than providing an out of date explanation that gets cut because of taskr hijacking the output.

### DIFF
--- a/src/commands/apm_cmds/publish.js
+++ b/src/commands/apm_cmds/publish.js
@@ -437,19 +437,15 @@ exports.task = function ({
             ctx.contract,
             from
           )
-          // Fix because APM.js gas comes with decimals and from doesn't work
+
           transaction.from = from
-          transaction.gas = Math.round(transaction.gas)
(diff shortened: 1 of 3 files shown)


src/commands/apm_cmds/publish.js
```diff
@@ -437,19 +437,15 @@ exports.task = function ({
...
 
           reporter.debug(JSON.stringify(transaction))
           
           return await web3.eth.sendTransaction(transaction)
         } catch (e) {
-          const errMsg = `${e}\nThis is usually one of these reasons, maybe:
-          - An existing version of this package was already deployed, try running 'aragon version' to bump it
-          - You are deploying a version higher than the one in the chain`
-          throw new Error(errMsg)
+          throw e
```
sohkai: We don't need this entire catch block at all, since we're just rethrowing it.

PR #233, opened 2018-10-19
Apm cleanup

ewingrj:
depends on https://github.com/aragon/aragon-cli/pull/230

Maybe we want to remove the `ipfs-rpc` flag from the cli and only support configuring this in the `arapp.json` environment `ipfsRPC` property


izqui:
Looking good, can you rebase with the current master?

0xGabi:
Rebase and iterate on #541

Issue #227, opened 2018-10-18
Some kits don't have their content published to ipfs-cache (aragen)


0x6431346e:
![image](https://user-images.githubusercontent.com/26041347/47156881-b7a79f80-d2f0-11e8-9ba1-5a5580c68c9e.png)
![image](https://user-images.githubusercontent.com/26041347/47156895-c0987100-d2f0-11e8-8b33-d87a327e7a8e.png)

To reproduce:
1. `aragon devchain`
2. `npm run start:local` on the `aragon/aragon` repo
3. Create a new organization

Alternatively:
1. `aragon init foo`
2. `aragon run`
3. Instead of going to your organization, go to the welcome screen and create a new organization

Related to https://github.com/aragon/aragon-cli/issues/177

izqui:
The current version of `aragen` doesn't have the DAO kits deployed, that's why the ENS names cannot be found. `aragen` versions will start containing these very soon.

sohkai:
@0x6431346e With v5.2.0, I think this should be fixed now? I think we can also close #177 since we couldn't reproduce it and it was a fairly long time ago.

0x6431346e:
I think we can close #177 since we cannot replicate it, but this one is not yet fixed, just tried it with `v5.2.0`

sohkai:
@0x6431346e are those related to #320?

This should be working with the latest aragen.

0x6431346e:
@sohkai Probably, but it should work at least with `aragon/aragon`, but now it gets:
`Error: Could not fetch ABI for template 'multisig'
    at Object.newDAO (index.js:104)`

In `aragon.js` we have this piece of code:
```
      const { contractAddress, abi } = await apm.getLatestVersion(tmplObj.appId)

      if (!contractAddress) {
        throw new Error(`No contract found on APM for template '${templateName}'`)
      }
      if (!abi) {
        throw new Error(`Could not fetch ABI for template '${templateName}'`)
      }
```
And in the CLI we have this workaround, in `dao new`:
```
          // TODO: Remove hack once https://github.com/aragon/aragen/pull/15 is finished and published
          const abi = ctx.repo.abi || BARE_KIT_ABI
```

Looking into it a bit more, I can see `apm versions` returns:
```
 % aragon apm versions --environment staging                                                 !11009
 ℹ multisig-kit.aragonpm.eth has 1 published versions
[...]

PR #238, opened 2018-10-22
Bump aragen to 2.0.0-beta.2 (with production kits)

izqui:
Bumps aragen to a recently generated version with the current state of aragonOS, apps and kits.

It is the first aragen release which contains the production kits `democracy` and `multisig`.

In order to use, fetch this branch and:

```
npm install
npm link
aragon devchain --reset # keep it running
aragon ipfs # keep it running
dao new --network rpc --kit [democracy-kit, multisig-kit or bare-kit] --fn newDemoInstance --fn-args [dao aragonid name]
```

After setting a name for a DAO, this name cannot be reused unless after a devchain reset.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -162,9 +162,9 @@
       }
     },
     "@aragon/aragen": {
-      "version": "2.0.0-beta.1",
-      "resolved": "https://registry.npmjs.org/@aragon/aragen/-/aragen-2.0.0-beta.1.tgz",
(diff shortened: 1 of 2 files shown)


0x6431346e:
This will resolve #227, is that correct?
I just tried to create a multisig and I get this:
```
Error: Could not fetch ABI for template 'multisig'
    at Object.newDAO (index.js:104)
``` 

Looks like `apm.getLatestVersion(''multisig-kit.aragonpm.eth')` is expected to return `{ contractAddress, abi }`

PR #246, opened 2018-10-23
import devchain from aragen

### LINKED ISSUE #264: aragon devchain does not show the output of ganache
Would be nice to see the transactions output, especially when debugging to see why a particular transaction failed.

ewingrj:
update cli to use https://github.com/aragon/aragen/pull/24

needs a release of aragen after the above PR is merged & to update `package.json`

Fixes #264.


0xGabi:
replace by #394

PR #223, opened 2018-10-16
Add silent and debug logging options

jvluso:
resolves #152

PR #249, opened 2018-10-23
enable more verbose logging in DEBUG mode

### LINKED ISSUE #125: Fix TaskList overwriting any console outputs during task execution
Right now it is almost impossible to debug the CLI due to TaskList overwriting them. We should improve current tasks so they can log messages.

perissology:
fixes #125 

if `DEBUG=true` we use the `verbose` listr render, which will not overwrite the console output


perissology:
yes, this can be closed. I should have checked the other pr first. I was thinking it added more logging output, not fixing the renderer (#125).

The only thing is we may want to add `dateformat: false` to the verbose renderer in that other PR. Seems a bit pointless to print the date to the console for the cli

0x6431346e:
Updated #223 accordingly

PR #253, opened 2018-10-25
add apm info & packages cmds

ewingrj:
adds a couple of cmds:

- `apm info` - to print the package information from apm
- `apm packages` - prints a list of all packages in the apm w/ their latest version


packages/aragon-cli/src/commands/apm_cmds/packages.js
```diff
@@ -0,0 +1,67 @@
...
+
+  apmOptions.ensRegistryAddress = apmOptions['ens-registry']
+  const apm = APM(web3, apmOptions)
+
+  const tasks = new TaskList([
+    {
+      title: `Fetching APM Registry: ${apmRegistry}`,
+      task: async (ctx, task) => {
+        // TODO add a new method to APM to allow fetching a registry without appId
+        ctx.registry = await apm.getRepoRegistry(`vault.${apmRegistry}`)
```
izqui: It would be safer to get some of the internal packages that APM deploys (like `apm-repo`)

packages/aragon-cli/src/commands/apm_cmds/packages.js
```diff
@@ -0,0 +1,67 @@
...
+      },
+    },
+    {
+      title: 'Gathering Repos',
+      task: async (ctx, task) => {
+        const e = await ctx.registry.getPastEvents('NewRepo', { fromBlock: 0 })
```
izqui: It would be better to get events just from the initialization block of the registry (which is exposed as the registry is an AragonApp)

2color:
Sweet! Nice addition @ewingrj

PR #254, opened 2018-10-26
allow setting permissions on app install

ewingrj:
allow setting the permissions via the cli when `dao install` is run.

Also changes the default permission manager for `open-permission` to `accounts[0]` instead of `ANY_ENTITY` to enable the permissions to created/revoked after installation. `ANY_ENTITY` is essentially burning the permission manager

### DIFF
--- a/src/commands/dao_cmds/install.js
+++ b/src/commands/dao_cmds/install.js
@@ -1,4 +1,6 @@
 const TaskList = require('listr')
+const taskInput = require('listr-input')
+const { isAddress } = require('web3-utils')
 const daoArg = require('./utils/daoArg')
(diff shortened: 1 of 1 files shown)


sohkai:
Nice find @perissology!

0x6431346e:
Oh man... I wish I was aware of this PR :cry:.  Would have solved #297

Related to #337

sohkai:
@0x6431346e Is this still relevant? I remember we had a discussion about permissions, their defaults, and providing a more interactive interface for assigning them.

0xGabi:
I do think this is still relevant

sohkai:
@0xGabi Were the changes in this PR still relevant?

PR #259, opened 2018-11-05
Pin IPFS to v0.4.17

### LINKED ISSUE #257: Unable to install @aragon/cli
On OSX 10.14, node 10.13.0 (managed via nvm 0.25.1)

Debug log:

https://gist.github.com/TalkingAntColony/cf18e69278a26df6524897b6744db8ce

tommycox:
Fixes #257
npm not being able to find IPFS during `npm i -g @aragon/cli`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7632,7 +7632,7 @@
       "resolved": "https://registry.npmjs.org/eth-sig-util/-/eth-sig-util-1.4.2.tgz",
       "integrity": "sha1-jZWCAsftuq6Dlwf7pvCf8ydgYhA=",
       "requires": {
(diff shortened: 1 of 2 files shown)


tommycox:
Something to consider: the tests didn't catch the failing case since the CLI was being installed from source and not npm... Perhaps the tests should utilize npm to ensure the same use cases that the end users are going to experience are covered.

izqui:
Woops sorry I missed the notification and already implemented it in #260 :(

Thanks so much for the contribution! Also saw your PR upstream, as soon as it gets released we will revert the change and pin to the fixed version.

Yeah, there are really no tests beyond the init command right now, it is something we need to work on.

PR #261, opened 2018-11-05
Add support for IPFS authorization headers

izqui:
Requires https://github.com/aragon/apm.js/pull/30

### DIFF
--- a/src/commands/ipfs.js
+++ b/src/commands/ipfs.js
@@ -14,25 +14,27 @@ exports.command = 'ipfs'
 exports.describe = 'Start IPFS daemon configured to work with Aragon'
 
 exports.task = ({ apmOptions }) => {
+  const ipfsOptions = apmOptions.ipfs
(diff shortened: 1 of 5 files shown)


src/commands/ipfs.js
```diff
@@ -14,33 +14,35 @@ exports.command = 'ipfs'
 exports.describe = 'Start IPFS daemon configured to work with Aragon'
 
 exports.task = ({ apmOptions }) => {
+  const ipfsOptions = apmOptions.ipfs
+  const ipfsRPC = `${ipfsOptions.protocol}://${ipfsOptions.host}:${ipfsOptions.port}`
```
0x6431346e: Looks like `ipfsRPC` is no longer needed

sohkai:
@0x6431346e I think the motivation for this PR came from the fact that `ipfs.eth.aragon.network`'s RPC API can be interacted with only through an authorization token. Does it make sense to update this functionality and pull it into the modern repo?

PR #262, opened 2018-11-06
Add ENS address output to `aragon devchain`

bingen:


src/commands/devchain.js
```diff
@@ -122,5 +122,6 @@ exports.handler = async ({ reporter, port, reset, accounts }) => {
   exports.printMnemonic(reporter, mnemonic)
   exports.printResetNotice(reporter, reset)
 
+  reporter.info("ENS instance deployed at 0x5f6f7e8cc7346a11ca2def8f827b7a0b612c56a1\n")
```
izqui: It is better if you get the address from `apmOptions['ens-registry']`, you would need to add it the option destructuring in L118

PR #270, opened 2018-11-12
Support 'dao upgrade' and 'dao install' on live DAOs (transaction pathing)

izqui:
Builds on #268 

A bit of a monster PR:
- Update to aragon.js v3.0.0
- Introduce a new optional environment argument (`wsRPC`) that if present is used by aragon.js as its data provider (as a workaround to truffle providers being on web3 0.x and not supporting web sockets). Mimics how the client has two different web3 instances, one for the 'wallet' (that signs transactions) and one for fetching events and data from the network.
- `dao install` and `dao upgrade` now use aragon.js for transaction pathing (e.g. start an upgrade that requires a vote in the organization).
- Adds support for using aragonid names and shortcuts when referring to DAOs in the CLI (`dao apps genesis` or `dao apps genesis.aragonpm.eth` are supported).


izqui:
Should be ready for review

src/commands/dao_cmds/install.js
```diff
@@ -38,72 +43,119 @@ exports.builder = function (yargs) {
...
         if (!ctx.repo.roles || ctx.repo.roles.length === 0) {
           throw new Error('You have no permissions defined in your arapp.json\nThis is required for your app to properly show up.')
         }
 
         const permissions = ctx.repo.roles
           .map((role) => [ANY_ENTITY, ctx.appAddress, role.bytes])
 
-        return setPermissions(
+        // TODO: setPermissions should use ACL functions with transaction pathing
+        return setPermissionsWithoutTransactionPathing(
```
izqui: `dao install` still doesn't set permissions using transaction pathing, so this will fail when the sender cannot set the permission directly. I think it should be its own issue though

0x6431346e:
Fixed the conflicts and gave it a good read :+1: , left to do is some manual testing

src/middleware/environment.js
```diff
@@ -52,9 +52,7 @@ const configureNetwork = (argv, network) => {
       provider = truffleNetwork.provider
     }
   } else if (truffleNetwork.host && truffleNetwork.port) {
-    provider = new Web3.providers.WebsocketProvider(
-      `ws://${truffleNetwork.host}:${truffleNetwork.port}`
-    )
+    provider = new Web3.providers.WebsocketProvider(`${truffleNetwork.host}`)
```
0x6431346e: Because of this change some commands don't work, e.g. `aragon apm versions`, unless I change my truffle.js to
```diff
- "host": "localhost",
+ "host": "ws://localhost:8545",
  "port": 8545
```
I think we can safely revert this though.

src/middleware/environment.js
```diff
@@ -117,13 +115,18 @@ module.exports = function environmentMiddleware(argv) {
...
     }
+
     if (env.registry) {
       resp.apmEnsRegistry = env.registry
       if (apm) {
         apm['ens-registry'] = env.registry
       }
     }
 
+    if (env.wsRPC) {
```
0x6431346e: I wanted to say this is a breaking change, since now you need the `wsRPC` property to do stuff like:
`dao apps 0xF9e4c6ee977b1Fe7D296FbE3b75BA05f12A30E4e --environment staging`
But this doesn't seem to work on master anyway, it gets stuck at _Inspecting DAO_.

1. Did dao commands ever work on something else than localhost ?
2. Is the wsRPC really necessary? I don't understand why the wrapper cannot fetch data using an http provider :confused:

It would have been nice to not need this extra `wsRPC` property, since I assume the cli doesn't need sandboxing like the dapp does.

galactusss: @0x6431346e Answering 2: To use the wallet-provider Jeremy builded we need to have `wsRPC` arg to pass the custom provider.

Here is the wallet-provider [Readme](https://github.com/jvluso/aragon.js/tree/wallet-provider-2/packages/aragon-wallet-provider)

src/middleware/environment.js
izqui: Totally. Did this for testing and forgot to revert

src/middleware/environment.js
izqui: In web3.js 1.0 only ws providers can listen to events unfortunately. So yeah, this has never worked unless using frame that has a ws provider

PR #277, opened 2018-11-21
Change defaults to vars in add files to ipfs

sebohe:
I looked at the 4th example of: https://github.com/ipfs/js-ipfs-api/blob/master/README.md#importing-the-module-and-usage

And noticed that the keys match so now it just passes the whole object.


sebohe:
Just wanted to comment that this change allows for unsafe behavior from users by using a remote ipfs API endpoint. So maybe it shouldn't get merged. Just my thoughts.

Issue #278, opened 2018-11-28
Add Git hook for linting


galactusss:
Start with ESLint and Prettier [#217](https://github.com/aragon/aragon.js/issues/217)

PR #279, opened 2018-11-29
Add husky for pre-commit: lint-stage & pre-push: test

galactusss:
Solve #278 

I checked several commands in a local environment and is working fine.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,24 @@
+{
+    "env": {
+      "browser": true,
+      "es6": true,
+      "node": true
+    },
+    "extends": [
+      "standard",
+      "standard-react",
+      "eslint-config-prettier",
+      "plugin:prettier/recommended",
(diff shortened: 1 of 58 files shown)


galactusss:
@0x6431346e Can you review this when you have time?

galactusss:
Great! Thank you for reviewing it Daniel. I'll take that into account for aragon.js

sohkai:
> Looks like standard and prettier conflicts in regards to comma-dangle and space-before-function-paren, but I guess it's better to keep on par with aragon/aragon.

I think the prettier plugin for eslint should resolve problems like this, if you add it as the last preset? But yes, best to do so and it'll also be easier to publish once we make our shared config available :).

0x6431346e:
Yes, it does fix it. I only mentioned it because now we're not 100% standard compliant, but I think the trade off is worth it :smile:

PR #280, opened 2018-12-01
Pin ganache version to 2.2.1

bingen:
`aragon run` fails with versions from 2.3.0.


0x6431346e:
I will merge this, so people building from source also have the fix

PR #288, opened 2018-12-07
Fix app install command

bingen:
Don't assign permissions on app installation, just check roles are
defined. Before permissions were created completely open (ANY_ENTITY)
but with a wrong manager so no change was possible.
To choose between this one, #287 or #289


0x6431346e:
This functionality was introduced in #270 as an optional flag: `--set-permissions false`. 
Available in the next release.

PR #304, opened 2018-12-20
Fix bug in `aragon dao acl` command

bingen:
It was failing with the error: "Cannot read property 'reduce' of undefined"


galactusss:
@bingen I wasn't able to reproduce. I tested it with react-kit boilerplate updated (https://github.com/aragon/aragon-react-kit-boilerplate/pull/21) and cli v5.2.2
Do you have a specific case I can try?

Issue #295, opened 2018-12-13
Token Manager Install Catch-22


0x6431346e:
from @lkngtn:

If you setup a local RPC instance and wish to deploy a new token manager into a DAO there is a catch-22. The token manager cannot be initialized unless there is a deployed minime contract with the token manager’s address set as its controller and you cannot deploy the token manager (and get its address) without initializing. 🤷‍♂️

**Allowing the user to deploy applications and then initialize them with a separate command would resolve this issue.**

Quazia:
An alternative to allowing users to deploy the application and initialize separately would be to provide a script option in the install command that would run some application-specific code inbetween the contract deploy and the application initialization.

lkngtn:
> initialize separately would be to provide a script option in the install command that would run some application-specific code inbetween the contract deploy and the application initialization.

I'm wondering what the advantage here would be as opposed to scripting the cli itself?

Quazia:
Yeah, separating it out is a better idea. Daniel suggested we just modify the install command with an option that prevents initialization.

0x6431346e:
> Allowing the user to deploy applications and then initialize them with a separate command would resolve this issue.

Looks like this is already possible with the `--app-init none` flag.

Installing a TokenManager on rinkeby:
1. Create a minime token: `dao token <name> <symbol> --environment staging` , depends on #335 
2. Install the app without initializing:
`dao install <dao> token-manager --environment staging --app-init none`
If you have the `Manage apps` permission it will get installed immediately: 
```
 ℹ Successfully executed: "Create a new upgradeable instance of 0x9fa3927f639745e587912d4b0fea7ef9013bf93fb907d29faeab57417ba6e1d4 app linked to the Kernel, setting its code to 0x8C06aEBF29F20A2e09b32F5d44cEa49Db3EC2eE0 ."
```
otherwise, if it goes to a vote first, you need to find the proxy address by going on etherscan to your app's Kernel and look for the `NewAppProxy` event.

3. Change the token controller to the new app's proxy address. (to be implemented)
4. Set some permissions to the new app to show up, otherwise `aragon.js` will not find it:
```
✖ No artifact found for 0x8C06aEBF29F20A2e09b32F5d44cEa49Db3EC2eE0
```
Example:
```
dao acl create <dao> <proxy-address> MINT_ROLE <entity> <manager>  --environment staging
```
5. Initialize the app
```
[...]

PR #312, opened 2018-12-31
Add help text to `aragon dao install --app-init` option

bingen:
If install command doesn't find initialization method, it won't
initialize the newly installed app, but this is not obvious from the
help instructions.
Although it's "dangerous", sometimes it's useful to install an app
without initializing. For instance, Token Manager: to initialize it
the token must be provided, and this token must have this new Token
Manager app as controller. Therefore token controller must be changed
in between creation and initialization of Token Manager app.

### DIFF
--- a/src/commands/dao_cmds/install.js
+++ b/src/commands/dao_cmds/install.js
@@ -44,7 +44,7 @@ exports.builder = function(yargs) {
     .args(daoArg(yargs))
     .option('app-init', {
       description:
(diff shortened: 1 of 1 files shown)


bingen:
See https://github.com/aragon/hack/pull/73

bingen:
See [this](https://github.com/aragon/aragon-cli/issues/295#issuecomment-453840792) for detailed instructions for the TokenManager case.

PR #325, opened 2019-01-05
Monorepo structure

### LINKED ISSUE #324: Create monorepo structure on aragon-cli
Include aragon-cli and [create-aragon-app](https://github.com/AragonDAC/create-aragon-app) as packages.

0xGabi:
Close #324


galactusss:
Not sure about:
1. This changes will break other packages that depends of aragon-cli
2. File .travis.yml need to be updated

sohkai:
> The way aragon.js does it, it's one root Readme, and the packages don't have one 🤔

This is something we should fix 😅 ; each package should have one

Issue #137, opened 2018-06-15
Reset devchain after updating the cli


izqui:
Given that updates to the cli can bring new kits or apps, the devchain should be restarted when we detect that the version has changed.

Quick way to do it is to append the cli version to the snapshot directory, which will also allow to go back to a previous snapshot when downgrading.

If we want to do a proper wipe every time, we could save a file in the `~/.aragon` directory (`~/.aragon/version`) where we write the version number every time devchain is started. If when starting the devchain, we detect the version has changed, we should force a `reset`.

0x6431346e:
Fixes: #214

PR #329, opened 2019-01-06
Create a new ganache db for each version

0x6431346e:
Fixes https://github.com/aragon/aragon-cli/issues/137

I opted for the quick way of appending the cli version to the snapshot directory instead of doing a full wipe because: it allows to go back to a previous snapshot when downgrading and the disk footprint seems negligible.

image

I also removed the port from the db name, was there an usecase for it? :thinking:


galactusss:
@0x6431346e I checked and I don't think there is any use case for that

PR #330, opened 2019-01-06
Handle arapp.json parse errors

### LINKED ISSUE #243: Handle arapp JSON parsing errors
After #230, if the `arapp.json` file is not a valid JSON file, the error that gets printed to the user is:

```
 ✖ Web3 cannot connect using the network provider.

Make sure 'aragon devchain' is running or your provider settings are correct.
```

We should catch the error parsing the arapp file and let the user know so it can be fixed.

0x6431346e:
Fixes #243 

Sample output:
```
% aragon run                                                                                
✖ Error parsing the aragon config file: arapp should have required property 'path', arapp.roles[0].params[0] should be string!
```


galactusss:
Tested! Great feature added

PR #332, opened 2019-01-09
Require artifacts when fetching published repo

mikec:
getRepoTask was ignoring the case where artifacts.json isn't found, which causes things to break further down the task queue (specifically, the 'Create DAO' task in run.js, which expects a `roles` array to exist in the repo). Removing the `artifactRequired: false` will make getRepoTask do the right thing and throw an error, giving a hint that you need to have IPFS or your local HTTP server running.


mikec:
@bingen this will make the CLI throw a more meaningful error when the HTTP server for the app isn't running

sohkai:
AFAIK I think we still want this behaviour for projects like aragon/aragon that don't get published with an artifact.json.

However, maybe we should only disable this check if `--only-content` is used cc @izqui @0x6431346e?

PR #334, opened 2019-01-11
Use tx pathing for setPermissions

### LINKED ISSUE #286: setPermissions on `dao install` should use transaction pathing
See https://github.com/aragon/aragon-cli/pull/270#discussion_r235126972

0x6431346e:
Closes #286 

This PR does not help if you don't have enough votes to reach quorum:
```
± % dao install showcase foo.open.aragonpm.eth --set-permissions open --environment staging   !10241
 ✔ Fetching foo.open.aragonpm.eth@latest
 ✔ Checking installed version
 ✔ Deploying app instance
 ↓ Fetching deployed app [skipped]
   → App wasn't deployed in transaction.
 ℹ Successfully executed: "Creates a vote to execute the desired action, and casts a support vote if possible"
 ⚠ After the app instance is created, you will need to assign permissions to it for it appear as an app in the DAO
```
After the vote passes you have to manually find the `NewAppProxy` event from the `Kernel` and call `dao acl create`.

### DIFF
--- a/src/commands/dao_cmds/install.js
+++ b/src/commands/dao_cmds/install.js
@@ -15,26 +15,6 @@ const listrOpts = require('../../helpers/listr-options')
 const addressesEqual = (a, b) => a.toLowerCase() === b.toLowerCase()
 const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
 
-const setPermissionsWithoutTransactionPathing = async (
-  web3,
-  sender,
-  aclAddress,
-  permissions
-) => {
-  const acl = new web3.eth.Contract(
-    getContract('@aragon/os', 'ACL').abi,
-    aclAddress
-  )
-  return Promise.all(
-    permissions.map(([who, where, what, manager]) =>
-      acl.methods.createPermission(who, where, what, manager).send({
-        from: sender,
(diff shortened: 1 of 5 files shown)


src/commands/dao_cmds/install.js
```diff
@@ -182,15 +184,21 @@ exports.task = async ({
...
+                execTask(dao, getTransactionPath, {
+                  reporter,
+                  apm: apmOptions,
+                  web3,
+                  wsProvider,
+                  silent,
+                  debug,
+                })
+                  // execTask returns a TaskList not a promise
+                  .then(tasks => tasks.run())
```
0x6431346e: The problem was `execTask` returns a `Listr.TaskList` not a promise, so we need to do `.run()` on the tasklist.

0x6431346e:
Just tried it with `dao install gabi vault --set-permissions open --environment aragon:rinkeby` and it works: installed the app and created a vote for permissions. :tada: 

You can also test locally with:
`dao new`
`dao install <dao> vault --set-permissions open`
This time a vote won't start because `0xb4124cEB3451635DAcedd11767f004d8a28c6eE7` has direct permissions (the `Create permissions` role).

Issue #296, opened 2018-12-13
Minime Deployment from CLI


0x6431346e:
from @lkngtn:

If you want to deploy a token manager app, it depends on having a Minime token. However, there is no option from the CLI in order to deploy this dependency. For a developer who is familiar with deploying arbitrary code this would be a minor annoyance as it would require them to deploy the Minime token separately–but for a power user who understands Aragon Apps but isn’t an experienced solidity developer this is a blocking issue.

I recommend providing an option for deploying minime token instances using the CLI.

Quazia:
I think this should look something like `aragon dao token <TOKENNAME> <DECIMALUNITS> <SYMBOL> <TRANSFERSENABLED> `. Not sure if it would be worth providing default values for some of these and making them optional.

PR #335, opened 2019-01-12
296 Minime Deployment from CLI

Quazia:
There's probably too much in the task right now, some of it should probably be split into utils, mainly like the two deploys but I'm not sure how re-usable they would be so didn't bother for now.

Might be a better way to handle the 'from' but this seems to work for the time being. Also, I'm not sure if there are any differences between the NPM minimetoken and the one in aragonOS but currently using the aragonOS one. I'm fairly sure they're the same.

Addresses issue #296

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -250,10 +250,9 @@
       }
     },
     "@aragon/apps-shared-minime": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@aragon/apps-shared-minime/-/apps-shared-minime-1.0.0.tgz",
-      "integrity": "sha512-DC0ylnEcsqED8eQIxriDbRTmSmqdUYWJrGmysI/UTuHvdSfsAy9awbx4Ea6To1Acd3F6kca5oPn6m9WjCAhlxQ==",
-      "dev": true
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@aragon/apps-shared-minime/-/apps-shared-minime-1.0.1.tgz",
+      "integrity": "sha512-DJAhj68hJLeZ6/KKKch9G3VAZT0SBoNN9JB6AETAy6M/Iiw0ZQZiCVltw8pJgeH2dvVlEC1y1XiG2rnG9RNhJA=="
     },
     "@aragon/apps-token-manager": {
       "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


src/commands/dao_cmds/token.js
```diff
@@ -0,0 +1,151 @@
...
+              0,
+              tokenName,
+              decimalUnits,
+              symbol,
+              transferEnabled,
+            ],
+          })
+          const gas = await deployTx.estimateGas()
+
+          const deployPromise = deployTx.send({ from, gas: gas + 1e6 })
```
0x6431346e: I'm not sure about `gas + 1e6`, will this work on mainnet, is it ok?

I was getting `The contract code couldn't be stored, please check your gas amount.`

Note that the estimated gas value was:
- on ganache, `3365974` for the factory and `3123814` for the token
- on rinkeby, `3365974` for the factory and `3089614` for the token

src/commands/dao_cmds/token.js
```diff
@@ -0,0 +1,131 @@
+const TaskList = require('listr')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const { getContract } = require('../../util')
+const listrOpts = require('../../helpers/listr-options')
+
+const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
+exports.command =
+  'token <token-name> <symbol> [decimal-units] [transfer-enabled]'
```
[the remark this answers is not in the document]
0x6431346e: On second thought, should we also put this under `aragon` instead of `dao` ?
Doing help on this command shows:
```
Positionals:
  dao         Address of the Kernel or AragonID)                        [string]
  token-name  Full name of the new Token                              [required]
  symbol      Symbol of the new Token                                 [required]
```
even though `dao token name symbol` works somehow :face_with_head_bandage:

src/commands/dao_cmds/token.js
```diff
@@ -0,0 +1,151 @@
...
+  // Decode sender
+  const accounts = await web3.eth.getAccounts()
+  const from = accounts[0]
+
+  return new TaskList(
+    [
+      {
+        title: 'Deploy the MiniMeTokenFactory contract',
+        task: async (ctx, task) => {
+          let artifact = getContract('@aragon/os', 'MiniMeTokenFactory')
```
sohkai: `@aragon/os@4.0.0` shouldn't have this contract anymore... you probably want [`@aragon/apps-shared-minime`](https://www.npmjs.com/package/@aragon/apps-shared-minime)

sohkai: Should now be fixed with `@aragon/apps-shared-minime@1.0.1` :).

src/commands/dao_cmds/token.js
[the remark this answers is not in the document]
0x6431346e: I guess `dao token new` it is :D

0x6431346e:
Should we merge this?

PR #339, opened 2019-01-13
New command: change-controller

0x6431346e:
Builds on top of #335 , waiting to get that merged first.

Not sure about the name of this command though, see [discussion](https://github.com/aragon/aragon-cli/pull/335#discussion_r247307710).

### DIFF
--- a/src/commands/dao_cmds/change-controller.js
+++ b/src/commands/dao_cmds/change-controller.js
@@ -1,0 +1,82 @@
+const TaskList = require('listr')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const { getContract } = require('../../util')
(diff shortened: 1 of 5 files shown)


src/commands/dao_cmds/change-controller.js
```diff
@@ -0,0 +1,82 @@
+const TaskList = require('listr')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const { getContract } = require('../../util')
+const listrOpts = require('../../helpers/listr-options')
+const chalk = require('chalk')
+
+exports.command = 'change-controller <token-address> <new-controller>'
```
izqui: Even more convinced now that these commands should be under a `dao token` or `dao minime` sub-command.

PR #341, opened 2019-01-16
Pin ganache to 2.2.1

### LINKED ISSUE #321: Cannot create dao with ganache-core@2.3
Doing `aragon dao new` on `master` gets stuck after sending the transaction:
```
 ✔ Fetching kit bare-kit.aragonpm.eth@latest
 ⠇ Create new DAO from kit
   Checking DAO
```
This works fine with `v5.2.2`.
Looking at the [diff](https://github.com/aragon/aragon-cli/compare/v5.2.2...master) I suspect is because of #302 . Maybe aragen needs to be rebuilt?

0x6431346e:
Pinning again ganache, because 2.3.2 causes a bug when creating a new dao, see #321.

Before we update ganache again we should test it will fix the `dao new` command.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3905,7 +3905,7 @@
     },
     "babelify": {
       "version": "7.3.0",
-      "resolved": "https://registry.npmjs.org/babelify/-/babelify-7.3.0.tgz",
+      "resolved": "http://registry.npmjs.org/babelify/-/babelify-7.3.0.tgz",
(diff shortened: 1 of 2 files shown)


galactusss:
Yes, I also think this is the way to go 👍

sohkai:
@0x6431346e I think you missed the `^` earlier 😄. Let's merge this if you'd like to release the CLI soon!

0x6431346e:
@sohkai Oops, nice catch :sweat_smile: I thought `npm i -S ganache-core@2.2.1` would do the trick

PR #342, opened 2019-01-17
Bump minor -> 5.3.0

0x6431346e:
The CI looks broken, do you think you can manually make a release @sohkai ?


sohkai:
@0x6431346e Should we also bump the aragon/aragon version (let's use 0.6.3 if we do :)).

sohkai:
@0x6431346e just pushed a commit so that it uses the tag version instead, so it's a bit more clear what version it's using :).

PR #352, opened 2019-01-22
fix(publish): fix fetching artifacts from previously deployed versions

sohkai:
We weren't passing the repo details through the task's context, so `copyCurrentApplicationArtifacts()` was always failing on trying to access the repo's `content`.

Note that testing this was a bit of an adventure, as it uses your local ipfs node by default (it can take a while for it to fetch some files). I've increased the ipfs timeout to 5min (rather than 10s) since it's likely better UX in the CLI to block for a few min than to immediately say it can't fetch the files.

Also removes some deprecated parameters in the publish task that weren't used anymore.


sohkai:
@0x6431346e Could you have a look at this? I move these changes to the new monorepo structure if they look good to you :).

0xGabi:
@sohkai I tested this one as part of #426. I notice that fetching the artifacts files in the `copyCurrentApplicationArtifacts` always fail ([particular line](https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/src/commands/apm_cmds/publish.js#L176)). Event when I publish an app and try again with the same app.

0xGabi:
Included in #426

PR #356, opened 2019-02-05
Implement `dao act` command for Agent app integration

izqui:
Requires https://github.com/aragon/aragon.js/pull/238 & https://github.com/aragon/radspec/pull/55 (link radspec -> aragon.js -> cli and make sure to `npm run build` everywhere)

Implements a new command `dao act` (which we should probably rename after the Actor -> Agent rename, but can't think of anything better) that provides some syntax sugar over `dao exec` for executing actions using Agent app instances in a DAO.

2019-02-12 at 10 49 28 am

## Testing

- Get the Agent app from https://github.com/aragon/aragon-apps/pull/580
- Get Aragon from https://github.com/aragon/aragon/pull/610 or https://github.com/aragon/aragon/pull/608
- `npm i && cd future-apps/agent && npm link @aragon/cli`
- `npx aragon run --client-path path/to/aragon`
- `npx dao apps [dao]`
- `npx dao act [agent proxy] [agent proxy or another contract]  "isForwarder()"`

### DIFF
--- a/src/commands/dao_cmds/act.js
+++ b/src/commands/dao_cmds/act.js
@@ -17,7 +17,7 @@ exports.builder = function(yargs) {
       type: 'string',
     })
     .positional('target', {
-      description: 'Function to be executed',
+      description: 'Address where the action is being executed',
       type: 'string',
     })
     .positional('signature', {
--- a/src/commands/dao_cmds/act.js
+++ b/src/commands/dao_cmds/act.js
@@ -3,6 +3,8 @@ const getAppKernel = require('./utils/app-kernel')
 const { ensureWeb3 } = require('../../helpers/web3-fallback')
 const ABI = require('web3-eth-abi')
 
+const EXECUTE_FUNCTION_NAME = 'execute'
+
 exports.command = 'act <agent-address> <target> <signature> [call-args..]'
 
 exports.describe = 'Executes a call from the Agent app'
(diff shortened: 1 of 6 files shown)


izqui:
This is ready for review now. A1 wants to release the Agent app this week with CLI support (we could release a beta?). Please ping me in aragon.chat if you have any issues.

src/commands/dao_cmds/act.js
```diff
@@ -0,0 +1,80 @@
...
+exports.describe = 'Executes an action from the Agent app'
+
+exports.builder = function(yargs) {
+  return yargs
+    .positional('agent-address', {
+      description: 'Address of the Agent app proxy',
+      type: 'string',
+    })
+    .positional('target', {
+      description: 'Function to be executed',
```
0xGabi: I think this description is a typo. Could be: "Address of the target contract"

0x6431346e:
```suggestion
      description: 'Address where the action is being executed',
```

sohkai:
@0xGabi Should we also remove the `package-lock.json` for now?

src/commands/dao_cmds/act.js
```diff
@@ -0,0 +1,80 @@
...
+}) {
+  const web3 = await ensureWeb3(network)
+  const dao = await getAppKernel(web3, agentAddress)
+
+  if (dao === ZERO_ADDR) {
+    throw new Error(
+      'Invalid Agent app address, cannot find Kernel reference in contract'
+    )
+  }
+
```
sohkai: What do you think about checking this is actually an Agent app via the app id?

0xGabi: Not sure if you asked me, but I think it is more clear to understand in that way

0x6431346e: Wouldn't we need to maintain a list of Agent apps, assuming there will be other releases in the future?

sohkai: They should all keep the same ID (`namehash('agent.aragonpm.eth')`), although perhaps we should just log a warning instead that you might not be interacting with the "official" agent app.

Other applications may use a similar `execute()` interface in the future 🤷‍♂️

Issue #358, opened 2019-02-09
E2E tests with sharness


0x6431346e:
The CLI can break unexpectedly due to external dependencies (like https://github.com/aragon/aragon-cli/issues/274 and https://github.com/aragon/aragon-cli/issues/215) or changes in the environment (node/os).

Library: [sharness](https://github.com/chriscool/sharness)
Examples: [ipfs-sharness-tests](https://github.com/ipfs/ipfs-sharness-tests)

Note: There is huge value in having this set up, and the most important commands like `create-aragon-app` and `aragon run` tested, but not so much in having everything tested, although it make sense to do it with the help of the community as these are really great first issues.

Commands to test:
- [x] aragon ipfs
- [ ] aragon devchain
- [ ] aragon run
- [x] create-aragon-app
- TBD

0xGabi:
After discussion in #363 we decide to migrate tests to node

PR #363, opened 2019-02-15
Testing: sharness environment

0xGabi:
Sharness setup and some documentation on how to write tests. 

We will add the first tests the next couple of days, we invite the community to help. The current commands we would like to test are listed in #358

### DIFF
--- a/test/sharness/aragon_ipfs.t
+++ b/test/sharness/aragon_ipfs.t
@@ -7,15 +7,13 @@ test_description="Test aragon ipfs command"
 test_launch_aragon_ipfs() {
   test_expect_success "'aragon ipfs' succeeds" '
     aragon ipfs > out_file 2> err_file &
-    PGIDDDDDD=$! &&
-    echo "Process Group ID $PGIDDDDDD"
+    PID=$! &&
+    echo "Process ID of the shell running this: $PID" &&
+    PGID=$(test_get_pgid_from_pid $PID) &&
+    echo "Process Group ID: $PGID"
   '
-  test_expect_success "'aragon ipfs' has a PGID" '
-    PGID=$! &&
-    echo "Process Group ID $PGID"
-  '
-
   # TODO Memory leak: tail never finishes
+  # TODO process substitution is not portable, only works in bash
   test_expect_success "'aragon ipfs' says all good" '
     timeout 180 grep -m 1 "running" <(tail -f out_file)
   '
@@ -27,10 +25,8 @@ test_kill_ipfs() {
   # -- is the default, which also means -15 or -TERM, sending the Termination signal
   # consider doing a graceful shutdown with -INT (-2) (interrupt - same as CTRL+C)
   # or using -KILL (-9) if a clean termination does not work
-  test_expect_success "'aragon ipfs' can be terminated" '
-   PGIDDD=$! &&
-   echo "Process Group ID $PGIDDD" &&
-   kill -- -$PGID
+  test_expect_failure "'aragon ipfs' can be terminated" '
+   kill -9 -$PGID
   '
 }
 
--- a/test/sharness/aragon_ipfs.t
+++ b/test/sharness/aragon_ipfs.t
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+test_description="Test aragon ipfs command"
+
+. ./lib/test-lib.sh
+
+test_launch_aragon_ipfs() {
+  test_expect_success "'aragon ipfs' succeeds" '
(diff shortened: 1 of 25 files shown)


test/sharness/aragon_ipfs.t
```diff
@@ -0,0 +1,37 @@
...
+  # TODO test_curl_resp_http_code "http://localhost:5001/api/v0/version" "HTTP/1.1 200 OK"
+  # TODO test check aragon-cache  "http://127.0.0.1:$API_PORT/ipfs/$HASH" "HTTP/1.1 200 OK"
+}
+
+test_kill_ipfs() {
+  # -- is the default, which also means -15 or -TERM, sending the Termination signal
+  # consider doing a graceful shutdown with -INT (-2) (interrupt - same as CTRL+C)
+  # or using -KILL (-9) if a clean termination does not work
+  test_expect_failure "'aragon ipfs' can be terminated" '
+   kill -9 -$PGID
```
0x6431346e: While learning `bash` is quite fun... should we reconsider using `sharness`?

A seemingly easy task like killing a background process: `aragon ipfs`, proves to be very time consuming. 👇 
After looking through ipfs' examples, the idea was simple:
```
> aragon ipfs &
> PID=$!
> kill $PID
```

The issue here was that `kill $PID` would only kill the shell process that invoked `node`, leaving `node` and `ipfs` running.
The solution is using `kill -- -$PGID` which is the syntax for killing a `process group id`. Luckily for us the `PID` and `PGID` are identical and retuned by `$!`:
image

Even though it works well from the terminal 🎉, with `sharness`, it proves a bit more difficult:

1. The `PID` is not identical to the `PGID`
2. The group of which the `PID` returned by `$!` is part of includes the `sharness` process, so not only it kills the `node`+`ipfs` processes, but also the test 😅 

image

What we could do is filter the node processes that are part of the `PGID`, but it feels like this should be easier (other tests too - not only this specific example).

1. Doesn't scale well - writing tests and maintaining them requires in-depth knowledge of shell scripting
2. Not very portable - some tricks like process substitution, used in making stuff work, are not POSIX-compatible

### How about we write these tests with with `node` and `ava`? @0xGabi @sohkai 

- Less effort: a contributor wouldn't need bash knowledge to write these tests, just javascript
- Easier to read (IMO)

`grep -m 1 "running" <(tail -f out_file)`
vs
```
fs.watch('./output', (eventType, filename) => {
  fs.readFile('./output', function (err, data) {
    if(data.includes('running')) {
      test.pass()
    }
  })
})
```
- We can leverage libraries. Example: use `web3.js` with `localhost:8545` to test that ganache is working correctly, make calls to contracts like ENS to check if they're setup correctly, etc.

I'm not `100%` sure, but I think we could do with `node` anything we would with `bash`.

sohkai: 0x6431346e Looking at these tests, it does seem much more maintainable if we just use node tooling rather than try and learn bash intricacies. Most people are not that well versed with shells, including those on the team, and this looks to be more of a nightmare to maintain than it would help 😅.

A few alternatives:

- Snapshot testing (e.g. https://github.com/kentcdodds/split-guide/blob/fb4b2a2ebc1fb8c3c010c2af1318861b8bb1bb13/src/bin/index.test.js#L16)
- Defining some custom runners (e.g. https://github.com/kentcdodds/split-guide/blob/fb4b2a2ebc1fb8c3c010c2af1318861b8bb1bb13/src/bin/index.test.js#L93 or https://medium.com/@zorrodg/integration-tests-on-node-js-cli-part-1-why-and-how-fa5b1ba552fe)
- Using something like [`nixt`](https://github.com/vesln/nixt) as a framework

0xGabi: Yes, make sense. @0x6431346e `ava` seems the easy choice here. @sohkai Is there a reason you chose that framework over other options? 

I would like us to consider `Jest` for aragonCLI cause test coverage right now is almost 0 and seems a really flexible framework to experiment with.

In the less likely case we want to stick with TAP there is [node-tap](https://www.node-tap.org) as well

sohkai: `nixt` supports other test runners; it simply might make some interactions with running CLI commands easier (instead of doing it ourselves like in the first two suggestions).

`jest` is nice as it supports snapshots.

sohkai: Found out that `ava` also supports snapshot testing, so FWIW we could just use `ava` 🤷‍♂️

0xGabi: Thanks Brett!

PR #359, opened 2019-02-09
[WIP] Upgrade @aragon/wrapper

0x6431346e:
`@aragon/wrapper` no longer exports ABIs

We need to upgrade after https://github.com/aragon/aragon.js/pull/236 is merged and released, in order to resolve https://github.com/aragon/aragon-cli/issues/344.

TODO:
- [x] fix unhandled promise rejections - by https://github.com/aragon/aragon.js/pull/238
```
 % dao apps showcase --all --environment aragon:rinkeby                                                                         
 ⠏ Inspecting DAO
   → Fetching apps for showcase.aragonid.eth...
   Fetching permissionless apps
(node:12872) UnhandledPromiseRejectionWarning: Error: No available storage method found.
    at /home/daniel/repos/aragon/aragon.js/packages/aragon-wrapper/node_modules/localforage/dist/localforage.js:2743:25
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
(node:12872) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 7)
(node:12872) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:12872) UnhandledPromiseRejectionWarning: Error: No available storage method found.
    at /home/daniel/repos/aragon/aragon.js/packages/aragon-wrapper/node_modules/localforage/dist/localforage.js:2743:25
 ✔ Inspecting DAO
 ✔ Fetching permissionless apps
 ✔ Successfully fetched DAO apps for 0x6c9ea46b5147fb185a8fd556bb47e37fd2b509e8

```

PR #371, opened 2019-02-19
aragon init prepare template for environments

0xGabi:


test/commands/init.js
```diff
@@ -48,5 +58,7 @@ test('prepare project template', async t => {
   t.falsy(await fs.pathExists(repoPath))
   t.is(undefined, packageJson.license)
   t.falsy(fs.pathExistsSync(licensePath))
-  t.is(appName, project.environments.default.appName)
+  t.is(`${appName}`, project.environments.default.appName)
+  t.is(`${appName}.open.aragonpm.eth`, project.environments.staging.appName)
+  t.is(`${appName}.open.aragonpm.eth`, project.environments.production.appName)
```
0x6431346e: Looks like `appName` for staging and prod will be `aragon-app.aragonpm.eth.open.aragonpm.eth`

src/lib/init/prepare-template.js
```diff
@@ -5,25 +5,18 @@ export async function prepareTemplate(basename, appName) {
...
-  delete arapp.appName
-  delete arapp.version
+  const defaultEnv = arapp.environments.default
+  const stagingEnv = arapp.environments.staging
+  const productionEnv = arapp.environments.production
+
+  defaultEnv.appName = appName
+  Object.assign(arapp.environments.default, defaultEnv)
+
+  stagingEnv.appName = stagingEnv.appName.replace(/^app/, appName)
```
0x6431346e: To avoid `stagingEnv.appName = 'aragon-app.aragonpm.eth.open.aragonpm.eth'`

we could have:
```diff
- stagingEnv.appName = stagingEnv.appName.replace(/^app/, appName)
+ stagingEnv.appName = stagingEnv.appName.replace(/^app/, basename)
```

this will work for the CLI, but break the test, because in the test we are passing a path: `./.tmp/aragon-app`, and not the basename.

I think what we can do instead is some parsing in `prepareTemplate` (get the basename from appName) and rename the param to `path`:

```js
function prepareTemplate(path, appName) {
   const basename = name.split('.')[0]
   //...
}
```

0xGabi: Yes, you are right. This is better

package.json
```diff
@@ -11,7 +11,7 @@
     "build": "npm run extract-roles && babel -d dist src --copy-files",
     "prepare": "npm run build",
     "test": "ava",
-    "lint": "eslint src test && documentation lint src test",
+    "lint": "eslint src test/commands && documentation lint src test/commands",
```
sohkai: Just noticed this, but it seems unnecessary to run `documentation lint` on the tests 😄

src/lib/init/prepare-template.js
```diff
@@ -1,34 +1,27 @@
...
+  const gitFolderPath = path.resolve(dir, '.git')
+  const licensePath = path.resolve(dir, 'LICENSE')
 
-  const gitFolderPath = path.resolve(basename, '.git')
-  const licensePath = path.resolve(basename, 'LICENSE')
-
-  const packageJsonPath = path.resolve(basename, 'package.json')
+  const packageJsonPath = path.resolve(dir, 'package.json')
   const packageJson = await fs.readJson(packageJsonPath)
   delete packageJson.license
```
sohkai: Just curious, should we move this to also be in `create-aragon-app` (really really want to start pushing that!)?

0xGabi: Yes! I already made the same changes in `create-aragon-app`. Seems its never the perfect time to merge the monorepo PR to finally have `create-aragon-app` in aragonCLI 😅Maybe after the beta release?

package.json
0xGabi: I'll include this change in https://github.com/aragon/aragon-cli/pull/359

src/lib/init/prepare-template.js
sohkai: After the next release (5.4) would be a great time!

PR #385, opened 2019-02-28
check for existing transactionPath in execHandler

mikec:
`tryFindTransactionPath` was sometimes being called after `sendTransaction` resolved, which would clear ctx.transactionPath causing the reporter output `ctx.transactionPath[0].description` to fail

fixes this error I was getting in the reporter output (after successful transaction) using the `aragon dao acl create` command:

```
Sending transaction [started]
→ Waiting for transaction to be mined...
Sending transaction [completed]
 ✖ Cannot read property 'description' of undefined
```

Seemed like the lowest impact fix .. but might be better to clean up the promise code to avoid a race condition like this


sohkai:
@stale Please keep this open 😄

PR #390, opened 2019-03-02
Port e2e tests to node

0x6431346e:
Todo:

- [x] Fix CI


packages/e2e-tests/src/util.js
```diff
@@ -0,0 +1,75 @@
...
+) {
+  return new Promise((resolve, reject) => {
+    // start the process
+    const subprocess = execa(cmd)
+    logger(cmd, 'spawned with PID: ', subprocess.pid)
+    let stdout = ''
+    let stderr = ''
+
+    // return this function so the process can be killed
+    const exit = () => new Promise((resolve, reject) => {
```
0x6431346e: The same issue we had with bash, doing `subprocess.kill()` would not stop a background process.

For example:
```
13388 ~ 'node: npm test'
  `- 9164 ~ 'node: ava'
    `- 9488 ~ 'node: execa'   <- this is what I get in the test, which does not want to die :(
      `- 11808 ~ 'node: aragon ipfs' 
        `- 24548 ~ 'ipfs' 
```

Even if we listen to the 'exit' event in the CLI:

```js
  process.on('exit', (code, signal) => {
    console.log('parent said exit', code, signal)
    ipfsProc.kill()
    process.exit()
  })
```

Most likely `child.pid` returns a Group ID, and node cannot kill groups on windows: https://github.com/nodejs/node/issues/3617

Solution: https://www.npmjs.com/package/ps-tree which returns the 2 children which we can kill. 🔪

packages/e2e-tests/src/cli/version.js
```diff
@@ -0,0 +1,47 @@
...
+nixt.prototype.promise = function() {
+  return new Promise((resolve,reject) => {
+    this.end(err => {
+      if(err) reject(err)
+      resolve()
+    })
+  })
+}
+
+test('should return the correct version', async t => {
```
0x6431346e: There are some issues with `nixt`:
- for some reason the callback style `.end()` function did not play nice with `ava` (could have been because of me), had to extend it to support promises.
- does not support snapshots - I don't think there is a way to return the `stdout` so we cannot use our own assertions.
- [does not support background tasks](https://github.com/vesln/nixt/issues/17), which is why I created the `startBackgroundProcess` utility.
- does not play nice with windows, commands need to be appended with `.cmd`, see [issue](https://github.com/nodejs/node/issues/3675#issuecomment-154264390), they should use `execa` or `cross-spawn`.

I think we should simply use `execa` and `ava` assertions instead.

packages/e2e-tests/src/util.js
0xGabi: Great solution 🎉

packages/e2e-tests/src/cli/version.js
0xGabi: Seems a great tool anyway, but those are big downside in our case. Maybe when we test a bigger command like `aragon run` we will discover if only `execa` and `ava` is enough or we need at some point add other tooling

0x6431346e:
Finally made Travis work!

First it was failing because of the lockfiles: [github:ahultgren/async-eventemitter](https://travis-ci.org/aragon/aragon-cli/builds/501454718#L470) and [github:frozeman/WebSocket-Node](https://travis-ci.org/aragon/aragon-cli/builds/501492149#L476)

Then with `ENOENT` even though the packages **should** have been linked by `lerna bootstrap` (I even tried `--force-local`): https://travis-ci.org/aragon/aragon-cli/builds/501944557#L579

Thirdly, because of OS specific text, see `src/util.js#normalizeOutput`.

Ended up using hacky solution: running `npm link` after installing 🤔

.gitignore
```diff
@@ -3,14 +3,17 @@
 
 # Dependencies
 node_modules
+*ipfs.cmd*
```
0x6431346e: This is created at the root folder when installing `go-ipfs` on windows 🤷‍♂️

.travis.yml
```diff
@@ -9,8 +9,8 @@ before_script:
 install:
   - npm install
 script:
+  - npm run link
   - npm run lint
   - npm run test
-  - npm run build
```
0x6431346e: The `build` script is automatically ran after install because of [npm prepare](https://docs.npmjs.com/misc/scripts)

sohkai:
On `npm link`, I don't know what the particular issue is, but apparently [`lerna bootstrap` is dying for `lerna link`](https://github.com/lerna/lerna/issues/1728#issuecomment-464877442) so maybe that's could be something to look into.

0x6431346e:
> Then with `ENOENT` even though the packages **should** have been linked by `lerna bootstrap` (I even tried `--force-local`): [travis-ci.org/aragon/aragon-cli/builds/501944557#L579](https://travis-ci.org/aragon/aragon-cli/builds/501944557#L579)

> On `npm link`, I don't know what the particular issue is, but apparently [`lerna bootstrap` is dying for `lerna link`](https://github.com/lerna/lerna/issues/1728#issuecomment-464877442) so maybe that's could be something to look into.

Looks like it was an issue on my part actually :sweat_smile:.
I described the issue here: https://github.com/aragon/aragon-cli/pull/439/files#diff-c3d8c26e209da78ecf8c501174996f65R18

PR #403, opened 2019-03-17
Add source-maps for better stack traces

0x6431346e:
Before:
```
 > TypeError: Cannot read property 'links' of undefined
    at _callee2$ (E:\repos\aragon\aragon-cli\packages\aragon-cli\dist\commands\ipfs_propagate.js:80:62)
```

After:
```
 > TypeError: Cannot read property 'links' of undefined
    at links (E:\repos\aragon\aragon-cli\packages\aragon-cli\src\commands\ipfs_propagate.js:37:56)
```


packages/aragon-cli/package.json
```diff
@@ -71,6 +73,7 @@
     "opn": "^5.3.0",
     "rimraf": "^2.6.2",
     "semver": "^5.4.1",
+    "source-map-support": "^0.5.11",
```
0x6431346e: I've tried using `@babel/register` instead, but it uses relative paths which my editor cannot open...:
```
 > TypeError: Cannot read property 'links' of undefined
    at console (../../src/commands/ipfs_propagate.js:37:11)
```

Note: `@babel/register` [uses `source-map-support` under the hood](https://github.com/babel/babel/blob/master/packages/babel-register/src/node.js#L16).

packages/aragon-cli/package.json
```diff
@@ -8,11 +8,13 @@
     "dao": "./aliases/dao"
   },
   "scripts": {
-    "build": "npm run extract-roles && babel -d dist src --copy-files",
+    "build": "npm run extract-roles && npm run compile:src",
+    "watch": "npm run build -- -- --watch",
```
0x6431346e:
```suggestion
    "start": "npm run build -- -- --watch",
```

sohkai: I think `start` feels kind of weird, since you're not technically starting anything (you'd still need to link it?). What about `build:watch`?

0x6431346e: Makes sense. Would be awesome if we could come up with a "community standard" like:
`npm start` - for production use (build, link, etc.)
`npm run dev` - for dev use

So you only one command (besides install) to get up and running, without looking in `package.json`.

Would've been cool if the `npm` team "reserved" a script for this. :D

0x6431346e:
```suggestion
    "build:watch": "npm run build -- -- --watch",
```

Issue #383, opened 2019-02-27
Don't stop ipfs until it has finished


0x6431346e:
can be easily tested by going between this commit and the one before it: https://github.com/AutarkLabs/planning-suite/commit/e3178ebd51c4ac948d7fb06ea4609fe78d506c7d.

_Originally posted by @sohkai in https://github.com/aragon/aragon-cli/pull/360_

PR #406, opened 2019-03-21
Add ipfs view & ipfs propagate

### LINKED ISSUE #384: new command: aragon ipfs propagate <hash>
See the discussion in https://github.com/aragon/hack/pull/95

0x6431346e:
Closes #384


Specification.md
```diff
@@ -1,17 +1,145 @@
...
+- installing & starting a daemon with the right configuration
+- interacting with local/remote nodes: authentication, configuration, pinning, etc.
+- help setup a "production" node (?)
+- automatically pin anything published to an APM repository (?)
+
+### IPFS Binaries
+
+Local node:
+
+- ✔️ `aragon ipfs` - Alias for `aragon ipfs start`
```
0x6431346e: I've added a bunch of potential changes here after brainstorming a bit about the `ipfs` and `devchain` commands. Would love some feedback! cc: @0xGabi @sohkai @izqui 

I think people usually "optimize" their process and run `ipfs` and `devchain` in a separate terminal, so that `aragon run` and other commands are faster. This is why I think we should not stop IPFS/Devchain as part of these commands, but instead expose commands to do so.

Example 
> IPFS not started, do you wanna start it? [yes/no] 
Remember to do `aragon ipfs stop` afterwards.

And that `ipfs start` should finish and leave the daemon running in the background.

sohkai: I agree, having management commands for the daemon would be great! It'd also solve things like the race condition in https://github.com/aragon/aragon-cli/issues/383.

Specification.md
```diff
@@ -1,17 +1,145 @@
...
 
-By default `web3.js` uses `web3.eth.Contract.method.myMethod().estimateGas()`, but this value can differ from the 
-actual gas that will be used, if the contract you are calling depends on the blockhash, blocknumber or any other source of 
-randomness making the gas cost nondeterministic.
+The CLI should package several "core" extensions.
+The API these extensions expose is aimed at `node` environments.
+The aim is to provide convenience to power-users & devs.
 
-E.g.: the [`MiniMeToken`](https://github.com/aragon/aragon-apps/blob/master/shared/minime/contracts/MiniMeToken.sol) contract which snapshots balances at certain block numbers.
+## IPFS `@aragon/ipfs-utils`
```
sohkai: For extensions related to the cli, perhaps it would make sense to prefix them with `cli` (so this would become `@aragon/cli-ipfs-utils`)?

Specification.md
```diff
@@ -1,17 +1,145 @@
...
 
-Considering the above-mentioned behavior, the CLI should calculate the recommended gas limit as follows:
-- `recommendedGas = estimatedGas`, if `estimatedGas > upperGasLimit` or
+Goals:
+
+- installing & starting a daemon with the right configuration
+- interacting with local/remote nodes: authentication, configuration, pinning, etc.
+- help setup a "production" node (?)
+- automatically pin anything published to an APM repository (?)
```
sohkai: I wouldn't really consider these last two part of the flow; for production you're likely to set up your own infra and the last item is very abusable (we actually used to do this but have since moved away to curating the pinning set).

Specification.md
```diff
@@ -1,17 +1,145 @@
...
+  ```
+
+  - Should save `stdout` and `stderr` to files, e.g.: `stdout-${number}.log` in the `logsLocation`
+  - Should warn if it's already running
+  - Should error (exit code 1) if it cannot start (missing libs/ports taken)
+  - Should inform about where the logs are saved and how to stop it
+  - Should run `aragon ipfs status` and finish afterwards (optional)
+- `aragon ipfs stop` - Stop the **IPFS Daemon**
+  - Should warn if the node was already stopped
+- `aragon ipfs enable-startup` - Start the **IPFS Daemon** automatically at start-up
```
sohkai: May also make sense to add the ability to clear the IPFS db. Is the plan for us to use our own DB (rather than one the user may already have from a previous IPFS installation)?

packages/aragon-cli/src/lib/ipfs/index.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+function isDir(data) {
+  return data.length === 2 && data.toString() === '\u0008\u0001'
+}
+
+function stringifyMerkleDAGNode(merkleDAG) {
+  // ${merkleDAG.isDir ? '📁' : ''}
+  const cid = merkleDAG.cid
+  const name = merkleDAG.name || 'root'
```
0xGabi: When viewing a `cid` that is a single file the name isn't display, only `root`. For example:
```
 aragon ipfs view QmNsZ1b2uXNEYKx17anDvRoJkvvmmeXovz7stKYDEbpRyW
 ✔ Connect to IPFS
 ✔ Fetch the links
─ root - 15.9kB - QmNsZ1b2uXNEYKx17anDvRoJkvvmmeXovz7stKYDEbpRyW
```
In this example the file is `artifact.json`. Ideally would be great to display the name.

Specification.md
0x6431346e: Hmm, makes sense.
What about `dao`, `apm`, etc? I think we can publish those without the `utils` suffix, e.g.: `@aragon/cli-dao`.

Specification.md
0x6431346e: > May also make sense to add the ability to clear the IPFS db.

👍 

> Is the plan for us to use our own DB (rather than one the user may already have from a previous IPFS installation)?

Hmm, we could totally do this, but also allow it to be the default directory. I wonder what should be the default.. 🤔

packages/aragon-cli/src/lib/ipfs/index.js
0x6431346e: AFAIK the name of the hash cannot be retrieved. A way to preserve it would be to "wrap" files in a directory, that way, the DAG of the wrapper has a link to the file and it's name.

I think the idea behind it is that I can upload a file called `test.json` with the exact same contents of `artifact.json` and it will give me the same hash: `QmNsZ1b2uXNEYKx17anDvRoJkvvmmeXovz7stKYDEbpRyW` so the data is not duplicated or some other reason 🤔

packages/aragon-cli/src/commands/run.js
```diff
@@ -6,7 +6,7 @@ const publish = require('./apm_cmds/publish')
 const devchain = require('./devchain')
 const deploy = require('./deploy')
 const newDAO = require('./dao_cmds/new')
-const startIPFS = require('./ipfs')
+const startIPFS = require('./ipfs_cmds/start')
```
0x6431346e: 😅 This one almost slipped... can't wait to have typescript or tests to catch these things.

Specification.md
0xGabi: I like `@aragon/cli-<package>` format

packages/aragon-cli/src/lib/ipfs/index.js
0xGabi: Oh, that's why 👌

PR #422, opened 2019-04-02
Fix staging and production references from arapp.json.

ajsantander:
While trying to use the template in [aragon-react-boilerplate](https://github.com/aragon/aragon-react-boilerplate), when executing the command `npx create-aragon-app app.aragonpm.eth react --debug`, I was getting the following error:

```
Preparing template [failed]
→ Cannot read property 'appName' of undefined
 ✖ Cannot read property 'appName' of undefined
 ❯ TypeError: Cannot read property 'appName' of undefined
    at _callee$ (/home/user_name/.npm/_npx/2243/lib/node_modules/create-aragon-app/dist/lib/prepare-template.js:42:45)
    at tryCatch (/home/user_name/.npm/_npx/2243/lib/node_modules/create-aragon-app/node_modules/regenerator-runtime/runtime.js:45:40)
...
```

This seems to be because [prepare-template.js is looking for "environments.staging" and "environments.production" entries in arapp.json](https://github.com/aragon/aragon-cli/blob/master/packages/create-aragon-app/src/lib/prepare-template.js#L10). However, `arapp.json` contains `environments["aragon:staging"]` and `environments["aragon:mainnet"]` instead, so `stagingEnv` and `productionEnv` both result undefined in `prepare-template.js`, and the command fails when trying to get `appName` of undefined.

My fix in this PR is just for illustrative purposes. I'm sure that there is a more correct way to fix this, probably by making a change in the template instead of the cli. However, I do confirm that the Counter app works perfectly with this small change.

### DIFF
--- a/packages/create-aragon-app/src/lib/prepare-template.js
+++ b/packages/create-aragon-app/src/lib/prepare-template.js
@@ -7,16 +7,16 @@ export async function prepareTemplate(dir, appName) {
   const arapp = await fs.readJson(arappPath)
 
   const defaultEnv = arapp.environments.default
-  const stagingEnv = arapp.environments.staging
-  const productionEnv = arapp.environments.production
+  const stagingEnv = arapp.environments["aragon:staging"]
+  const productionEnv = arapp.environments["aragon:mainnet"]
 
   defaultEnv.appName = appName
   stagingEnv.appName = stagingEnv.appName.replace(/^app/, basename)
   productionEnv.appName = productionEnv.appName.replace(/^app/, basename)
 
   Object.assign(arapp.environments.default, defaultEnv)
-  Object.assign(arapp.environments.staging, stagingEnv)
-  Object.assign(arapp.environments.production, productionEnv)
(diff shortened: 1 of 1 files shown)


0x6431346e:
Thanks a lot for catching this!

I've reverted the change on the template: https://github.com/aragon/aragon-react-boilerplate/pull/49 so we don't have to release a new cli version asap.

I'm thinking this logic in `prepare-template.js` should be changed as it's not very flexible.

Perhaps it could be something like scan all files and replace `appName` with `foo`, rather than coding each case ourselves, manually.

ajsantander:
That was super fast! Thx ^^

sohkai:
@0x6431346e That makes sense, a lot of other templating tools use _very special_ strings for things like this 👍

0x6431346e:
Closing in favor of https://github.com/aragon/aragon-cli/pull/424
Thanks again @ajsantander 🙏 

> @0x6431346e That makes sense, a lot of other templating tools use _very special_ strings for things like this 👍

I couldn't think of anything better than `placeholder-app-name` 😄

Issue #133, opened 2018-06-12
solidity-extractor is only run on the main contract file, not the entire inheritance tree


sohkai:
See https://github.com/aragon/aragon.js/issues/118 for an example where this is a problem. Ideally we'd look through the entire inheritance tree for external methods to add to the artifact.json.

izqui:
I was about to start working on an improved extractor and found a PR adding this to truffle, which would make notice strings available in truffle build artifacts https://github.com/trufflesuite/truffle/pull/1221

Still thinking about whether it would make sense to have a standalone extractor that wouldn't depend on truffle, which would just extract all the notice strings from a contract.

sohkai:
> Still thinking about whether it would make sense to have a standalone extractor that wouldn't depend on truffle

I think the most robust way of extracting this info would just be compiling a contract and extracting the natspec from the solc output. However, doing this compilation requires tools like truffle for resolving files, etc.

izqui:
Yep, that's what I meant, having a extractor that would use solc to get the `userdoc`. My question was whether we should build that as a standalone tool or depend on truffle for that.

sohkai:
Given that we currently bundle truffle and use it to compile / test / etc, I think we should initially just use truffle's artifact output and add support for other methods when/if someone needs that functionality.

izqui:
Truffle has already merged it into their `next` branch, so it will be released with version 5

Quazia:
Moving to close assuming changes to truffle resolve this issue.

sohkai:
Looks like this was done with truffle@5, so it would be a good reason to upgrade to it :+1:.

0xGabi:
Related to #395

0xGabi:
related to #719

0xGabi:
Now that we update to truffle 5 I'll close this one

sohkai:
@0xGabi Just wanted to confirm, are we using the AST when the project is using truffle@5 now?

PR #423, opened 2019-04-02
Implement 'aragon apm extract-functions' command

izqui:
This new command exposes the `solidity-extractor` helper to extract function information (roles and radspec notice strings mainly) of a solidity file and generate minimal artifact files. Useful for generating artifacts for system apps that aren't published to APM through the proper process [like these](https://github.com/aragon/aragon.js/tree/master/packages/aragon-wrapper/artifacts/aragon).

I am not totally convinced about the name of the command (`extract-functions`) nor having it under the `apm` subcommands, suggestions are welcome. 

TODO: 

- [x] Add documentation


sohkai:
Would be nice if this also let you parse inherited contracts, as a way of fixing https://github.com/aragon/aragon-cli/issues/133.

PR #426, opened 2019-04-05
Sanity check artifacts in apm publish

### LINKED ISSUE #357: Reusing artifacts on apm publish should sanity check artifact
It should ensure that the network, repo name, as well as contract path are still correct and warn the user that it can't reuse an existing one due to mismatches.

### LINKED ISSUE #450: On major deployments, add `methods` from old versions' artifact.jsons
This is now an issue with our Finance app, where we used to have a method named `newPayment()` but has since been split out into two methods (`newScheduledPayment()` and `newImmediatePayment()`).

It would be nice if the CLI automatically consulted old versions' `artifact.jsons and included them as part of a versioned list of old methods, so radspec descriptions can still work on breaking upgrades.

0xGabi:
closes #357, closes #450

Iterate on #462 and build on #352


packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js
```diff
@@ -0,0 +1,215 @@
...
+  // > [{ sig: 'transfer(address)', role: 'X_ROLE', notice: 'Transfers..'}]
+  artifact.functions = await extract(path.resolve(cwd, artifact.path))
+  // extract abi for each function
+  // > [{ sig: , role: , notice: , abi: }]
+  decorateFunctionsWithAbi(artifact.functions, artifact.abi, web3)
+
+  // Consult old (major) version's artifacts and return an array
+  // of deprecated functions per version
+  // > "deprecated": { "1.0.0": [{}], "2.0.0": [{}] }
+  artifact.deprecated = await deprecatedFunctions(apm, artifact, web3, reporter)
```
sohkai: @0xGabi What do you think about making this `deprecatedFunctions` instead? `deprecated` sounds a bit vague (and might be taken to meant that the entire app is deprecated).

packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js
```diff
@@ -0,0 +1,215 @@
...
+        web3.eth.abi.encodeFunctionSignature(f.sig)
+    )
+  })
+}
+
+async function deprecatedFunctions(apm, artifact, web3, reporter) {
+  let deprecated = {}
+  try {
+    const deprecatedFunctionsSig = new Set()
+    const versions = await apm.getAllVersions(artifact.appName)
```
sohkai: Didn't notice this before, but we probably want to do some error handling and optimization around this.

`apm.getAllVersions()` is quite expensive; it'll fetch the IPFS content for each version. An optimization we may want to make is to not use `apm.js`, but something like https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/core/apm/repo.js#L9 to extract all the on-chain information first, and only load the IPFS content we actually need (the last version of each major range).

We should also do a bit of error handling, where if this fails, we prompt the user if they'd like to continue anyway. Not all upgrades will have deprecations, and it'd be nice if the user could choose to proceed to publishing even if they couldn't fetch the last published version's content.

0xGabi: Make sense. I'll take the opportunity and also move `generate artifact` task to their own command. I think is better to separate the concerns here and not having it in the `publish` command

Issue #417, opened 2019-03-25
Testing roadmap


0xGabi:
Steady process with the objective of improve testing across the repo. The tests will follow the [testing documentation](https://github.com/aragon/aragon-cli/blob/master/docs-internal/Testing.md). Below is a tentative roadmap.

- [x] `create-aragon-app`

- [x] `ipfs-cmds` #415

- [ ] `utils-cmds` (devchain, deploy and contracts)

- [x] `apm-cmds`

- [ ] `dao-cmds`

  - [ ] `token-cmds`

  - [ ] `acl-cmds`
 
- [x] `utils-cmds` (run)

0xGabi:
Closed for #602

PR #432, opened 2019-04-11
Testing strategy

0xGabi:
## Proposed changes

Add new testing documentation and standard for the repository. Will be used to work on #417


TESTING.md
```diff
@@ -0,0 +1,71 @@
...
+## Testing environment
+
+The testing environment consist of the following tools:
+
+1. Structure and syntax -> [AVA framework](https://github.com/avajs/ava)
+2. Assertion functions -> [AVA assertions](https://github.com/avajs/ava/blob/master/docs/03-assertions.md)
+3. Mocks, spies, and stubs -> [Sinon.JS](https://sinonjs.org)
+4. Code coverage -> [Istanbul](https://istanbul.js.org) with [nyc](https://github.com/istanbuljs/nyc) wrapper. We also use [Coveralls](https://coveralls.io) for adding coverage reporting to [Travis CI](https://travis-ci.org) flow.
+
+**Note:** In some instances you might found useful to use [proxyquire](https://github.com/thlorenz/proxyquire) to mock an Aragon wrapper.
```
0x6431346e: I think this applies only to `aragon.js` 🤔, maybe have this instead:
```suggestion
**Note:** In some instances you might found useful to use [proxyquire](https://github.com/thlorenz/proxyquire) to mock dependencies that should not be evaluated.
```

TESTING.md
```diff
@@ -0,0 +1,71 @@
+# Testing
+
+We have two different processes. One for running unit and integration tests and another one for functional tests. This is because functional tests usually take longer.
```
0x6431346e:
```suggestion
We have two different processes. One for running unit and integration tests and another one for functional tests (also called end-to-end tests). This is because functional tests usually take longer.
```

> unit tests
> integration tests
> functional tests

Let's add some links or definitions to these concepts, e.g.:

https://en.wikipedia.org/wiki/Unit_testing
https://en.wikipedia.org/wiki/Integration_testing
https://en.wikipedia.org/wiki/Functional_testing

Maybe we could include a picture like this pyramid:
https://medium.com/@Colin_But/define-testing-strategy-using-the-testing-pyramid-1dabee37e823

TESTING.md
```diff
@@ -0,0 +1,71 @@
...
+### Running tests
+
+#### Scripts
+
+Each package comes with the following scripts to run unit and integration tests:
+
+```json
+{
+  "test": "nyc ava",
+  "watch": "ava --watch"
```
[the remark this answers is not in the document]
0xGabi: Great, that was my first option but then I changed it cause we had it like that on the `e2e-tests` package. I'll change all the `package.json` files as well.

PR #433, opened 2019-04-12
Update PULL_REQUEST_TEMPLATE.md

0x6431346e:
I think we should aim to welcome any contributions regardless if they follow a specific PR template, read the guidelines of our contributing document or not.

With this change I've tried to lower the amount of text someone has to read and delete (by adding these instructions as comments).

I removed the "types of changes" section which I think might be better suited for us the maintainers to asses (since only us can add labels anyway) and some checkboxes that felt redundant.

We could keep this template short & sweet and the detailed instructions in the contributing doc.


0xGabi:
Totally agree!

PR #434, opened 2019-04-12
Update README.md

0x6431346e:
The docs in this file were very outdated and the shields were pointing to `aragon-dev-cli`.


sohkai:
Might be good to sanitize https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/README.md as well, I believe it still has the old badge too 😄

PR #455, opened 2019-05-04
Handle Aragon gateway on environment middleware

0xGabi:
# 🦅 Pull Request


0x6431346e:
Let's add a comment in the code on how this magic works :smile: 

E.g.: 

We check for the `registry` property on the environment to determine whether the environment is localhost or not. When we are not running locally we change the default ipfs rpc to the aragon gateway.

PR #462, opened 2019-05-15
Deprecated functions 

### LINKED ISSUE #450: On major deployments, add `methods` from old versions' artifact.jsons

0xGabi:
# 🦅 Pull Request

closes #450 

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [ ] Added/updated the necessary documentation
- [ ] Cleared dependencies on other modules that have to be released before merging (update Radspec)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,21 +3,129 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@lerna/add": {
-      "version": "3.14.0",
-      "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.14.0.tgz",
-      "integrity": "sha512-Sa79Ju6HqF3heSVpBiYPNrGtuS56U/jMzVq4CcVvhNwB34USLrzJJncGFVcfnuUvsjKeFJv+jHxUycHeRE8XYw==",
+    "@evocateur/libnpmaccess": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/@evocateur/libnpmaccess/-/libnpmaccess-3.1.0.tgz",
+      "integrity": "sha512-bfrqZ0v+Il5TJBsgF2oyepeJg34K2pBItapzP+UT1QMIGpUh/Zc1pQql4jrafamZTqP3ZvdJxaElat8B5K3ICA==",
       "dev": true,
       "requires": {
-        "@lerna/bootstrap": "3.14.0",
-        "@lerna/command": "3.14.0",
-        "@lerna/filter-options": "3.14.0",
+        "@evocateur/npm-registry-fetch": "^3.9.1",
+        "aproba": "^2.0.0",
+        "figgy-pudding": "^3.5.1",
+        "get-stream": "^4.0.0",
+        "npm-package-arg": "^6.1.0"
+      },
+      "dependencies": {
+        "aproba": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/aproba/-/aproba-2.0.0.tgz",
+          "integrity": "sha512-lYe4Gx7QT+MKGbDsA+Z+he/Wtef0BiwDOlK/XkBrdfsh9J/jPPXbX0tE9x9cl27Tmu5gg3QUbUrQYa/y+KOHPQ==",
+          "dev": true
+        }
+      }
+    },
+    "@evocateur/libnpmpublish": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@evocateur/libnpmpublish/-/libnpmpublish-1.2.0.tgz",
+      "integrity": "sha512-sezhX9FSnPIyrBBvxVocVJVO1uIWPczf6rOmUZSntCWfQMraO8pWTFlDJbroFqPbEqFFHf3eyw8NQ0Eb7OLd1g==",
+      "dev": true,
+      "requires": {
+        "@evocateur/npm-registry-fetch": "^3.9.1",
+        "aproba": "^2.0.0",
+        "figgy-pudding": "^3.5.1",
+        "get-stream": "^4.0.0",
+        "lodash.clonedeep": "^4.5.0",
+        "normalize-package-data": "^2.4.0",
+        "npm-package-arg": "^6.1.0",
+        "semver": "^5.5.1",
+        "ssri": "^6.0.1"
+      },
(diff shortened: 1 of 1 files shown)


packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+          if (
+            !artifact.functions.some(obj => obj.sig === f.sig) &&
+            !deprecatedFunctionsSig.includes(f.sig)
+          ) {
+            deprecatedOnVersion.push(f)
+            deprecatedFunctionsSig.push(f.sig)
+          }
+        })
+        if (deprecatedOnVersion.length) {
+          deprecated[`${lastMajor}.0.0`] = deprecatedOnVersion
```
0xGabi: @sohkai or @0x6431346e can you help with this line. I been struggling with JSON and how to add the array of deprecated functions. I tested the content of deprecatedOnVersionand is correct in this line, but is never assigned to deprecated (deprecated remain always [ ]). I also tested a simpler example on the console of the browser and it work like this 😔

sohkai: @0xGabi Unfortunately you can't assigned named keys to an array in JS, you'll have to use an object (`{}`) instead.

We can iterate over an object quite easily using helpers like `Object.entries()`, `for/of`, and etc. so don't worry about making `deprecated` an array :).

0xGabi:
Ready for review @sohkai

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+function getFunctionsAbi(functions, abi) {
+  functions.forEach(f => {
+    f.abi = abi.find(method => method.name === f.sig.match(/(.*)\(/)[1])
+  })
+}
+
+async function getDeprecatedFunctions(apm, artifact) {
+  let deprecated = {}
+  try {
+    let deprecatedOnVersion = []
```
sohkai: Let's pull this into the `forEach()` as an inner variable (so we don't have to reset it manually).

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+          }
+        })
+        if (deprecatedOnVersion.length) {
+          deprecated[`${lastMajor}.0.0`] = deprecatedOnVersion
+          getFunctionsAbi(deprecatedOnVersion, repo.abi)
+          deprecatedOnVersion = []
+        }
+      }
+    })
+    return deprecated
```
sohkai: I think it'd be good to move this return to the end of the function, so this function always returns an object.

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+  try {
+    let deprecatedOnVersion = []
+    const deprecatedFunctionsSig = []
+    const versions = await apm.getAllVersions(artifact.appName)
+    let lastMajor = -1
+    versions.reverse().forEach(async repo => {
+      // iterate on major versions
+      if (getMajor(repo.version) !== lastMajor) {
+        lastMajor = getMajor(repo.version)
+        repo.functions.forEach(f => {
```
sohkai: We should be careful here; if the IPFS content was lost for some reason, we wouldn't have the `repo.function` or `repo.abi`.

Maybe we could print a warning with this, and return early.

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+}
+
+async function getDeprecatedFunctions(apm, artifact) {
+  let deprecated = {}
+  try {
+    let deprecatedOnVersion = []
+    const deprecatedFunctionsSig = []
+    const versions = await apm.getAllVersions(artifact.appName)
+    let lastMajor = -1
+    versions.reverse().forEach(async repo => {
```
sohkai: Can we rename `repo` to `version` (or `repoVersion`)? It's a bit weird calling `version.version` but it is what it is :).

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -107,8 +107,50 @@ exports.builder = function(yargs) {
...
+  functions.forEach(f => {
+    f.abi = abi.find(method => method.name === f.sig.match(/(.*)\(/)[1])
+  })
+}
+
+async function getDeprecatedFunctions(apm, artifact) {
+  let deprecated = {}
+  try {
+    let deprecatedOnVersion = []
+    const deprecatedFunctionsSig = []
```
sohkai: This would be better as a JS `Set`.

packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -369,7 +418,7 @@ exports.task = function({
 
           if (!isValid) {
             throw new Error(
-              'Version bump is not valid, you have to respect APM bumps policy. Check version upgrade rules in documentation https://hack.aragon.org/docs/aragonos-ref.html#631-version-upgrade-rules'
+              'Version bump is not valid, you have to respect APM bumps policy. Check version upgrade rules in documentation https://hack.aragon.org/docs/apm-ref#version-upgrade-rules'
```
sohkai:
```suggestion
              'Version bump is not valid, you have to respect aragonPM bumps policy. Check version upgrade rules in documentation https://hack.aragon.org/docs/apm-ref#version-upgrade-rules'
```

If there are other places where we still use `APM` in a user-seen output, let's also change them to be `aragonPM` :).

0xGabi:
@0x6431346e this is ready to review 🤘

0xGabi:
Close over #426

PR #481, opened 2019-05-22
Include IPFS timeout and Aragon node as default for environments

0xGabi:
# 🦅 Pull Request

## ✔️ PR Todo

- [ ] Included links to related issues/PRs
- [ ] Added/updated unit tests for this change
- [ ] Added/updated the necessary documentation
- [ ] Cleared dependencies on other modules that have to be released before merging this


packages/aragon-cli/src/commands/dao_cmds/utils/getRepoTask.js
```diff
@@ -1,4 +1,6 @@
+const pkg = require('../../../../package.json')
 const LATEST_VERSION = 'latest'
+const DEFAULT_IPFS_TIMEOUT = pkg.aragon.defaultIpfsTimeouts
```
0x6431346e:
```suggestion
const DEFAULT_IPFS_TIMEOUT = pkg.aragon.defaultIpfsTimeout
```

PR #489, opened 2019-05-29
Feature to link deployed library contracts

### LINKED ISSUE #452: Allow declaring external contract dependencies
From the chat:

mike.calvanese 7:59 PM
>I added a library to our aragon app contract, and now getting an error trying to deploy ... `The data field must be HEX encoded data`
wondering if it's something similar to this issue in zos with library linking https://github.com/zeppelinos/zos/issues/751
the contract I'm trying to deploy is https://github.com/levelkdev/futarchy-app/blob/aragon-upgrades/contracts/Futarchy.sol

mike.calvanese 10:09 PM
>I implemented a workaround https://github.com/levelkdev/futarchy-app/commit/9a97599ca218167d24f9f34b257c5990f658a0de but definitely not ideal. is there support for something like truffle's `link()` in aragon that I'm missing?

sohkai Established Member 11:17 AM
>Hmm indeed, we definitely don't handle libraries well in aragonCLI cc @0x6431346e  @0xGabi . We could perhaps have a "linking" mapping in the arapp.json's environments, since each environment would have a different library address. You'd still have to deploy the library yourself via a script, but afterwards it'd be more seamless.

cslarson:
# 🦅 Pull Request

close #452

This feature links any linkable contracts into the deploying contract bytecode. It looks for these linkable contracts in a `links` array in arapp.json -> environments. Each object in that array should have `name` and `address` attributes.

Also, see related [issue](https://github.com/aragon/aragon-cli/issues/452#issuecomment-497109996)

## ✔️ PR Todo


packages/aragon-cli/src/commands/deploy.js
```diff
@@ -101,6 +103,13 @@ exports.task = async ({
 
           task.output = `Deploying '${contractName}' to network`
 
+          let environment = Object.values(module.environments)[0]
```
0x6431346e: Won't this work only with the first environment?

I think we need to do something like: 

```suggestion
          let environment = Object.values(module.environments)[environment]
```

packages/aragon-cli/src/commands/deploy.js
```diff
@@ -36,6 +36,7 @@ exports.builder = yargs => {
 }
 
 exports.task = async ({
+  module,
```
0x6431346e:
```suggestion
  module,
  environment,
```

(see my other comment)

cslarson: I agree, but it seems like only the selected environment is included in `module.envionments` by the middleware. See [here](https://github.com/aragon/aragon-cli/blob/master/packages/aragon-cli/src/middleware/environment.js#L113). Even if that's the case though I agree being explicit here is better.

0x6431346e: Oh ok, makes sense

cslarson: How about adding the selected environment as `module.env = env` in the environments middleware?

sohkai:
@0xGabi ideally we'd be able to test this before the actual deployment, as otherwise someone may have wasted ETH deploying a contract that wasn't linked properly.

We should be able to parse the compile output (e.g. in the `build/` folder), but I'd have to check a bit more deeply into the publish flow to see exactly when we compile and deploy the contract. At least one way of doing this would be to compare the contract bytecode to see if the expected linked library address was included or not (this could also be done after deployment, via [`web3.eth.getCode()`](https://web3js.readthedocs.io/en/1.0/web3-eth.html#getcode).

Not sure if there's a better way.

0xGabi:
Superseded by #804

PR #490, opened 2019-05-30
Fix: create-aragon-app

0xGabi:


0x6431346e:
Nice!

Please remember to update the https://github.com/aragon/aragon-cli/blob/master/docs-internal/Dependencies.md document. I still don't understand what is the problem with yargs 13 :man_shrugging:

PR #519, opened 2019-06-15
Fix a bunch of stuff

### LINKED ISSUE #445: Split ipfs rpc options
## 🚀 Feature

Split the `apm.ipfs.rpc` option into two:
- Allow specifying an ipfs gateway to read from
- Allow specifying an ipfs node to write to

## Motivation

Power users interacting with their DAO on rinkeby or mainnet have trouble finding the artifacts for these apps. That's because the CLI uses the `localhost` node by default, which does not bundle ALL the versions of the apps. See https://github.com/aragon/aragon-cli/issues/419

Currently, specifying the aragon ipfs node helps:
```
aragon apm info voting --environment aragon:mainnet --apm.ipfs.rpc https://ipfs.eth.aragon.network/ipfs 
```

This works, surprisingly, but only on some commands, because `apm.js` uses an old version of the `ipfs-api` package.
Note: some commands don't even pass along the apmOptions to the wrapper (only the ens option), e.g.: `dao apps`.

This shouldn't work at all because `https://ipfs.eth.aragon.network/ipfs` does not expose the API (port `5001`), but only the gateway (port `8080`), for obvious reasons.

## Pitch

1. We should do the reading only from the gateway option.
2. It should default to `localhost:5001` and `localhost:8080`, so that `aragon run` and other commands that rely on local assets would still work as expected.

As a second iteration we could include these options in `environments` 🤔.

### LINKED ISSUE #514: Not able to start IPFS daemon
## 🐛 Bug Report

```
>aragon ipfs start
  ✖ Start IPFS
    → Unexpected token p in JSON at position 4
```

## Context

I'm using last `master` branch (version `5.9.6`)
Logging `apmOptions.ipfs.rpc` I see that it has:

```
{ protocol: 'https',
  host: 'ipfs.eth.aragon.network',
  port: null }
```
which seems different from what `--help` says:
```
  --apm.ipfs.rpc      An URI to the IPFS node used to publish files
                                      [default: "http://localhost:5001#default"]
```

If I run this instead:
```
> aragon ipfs start --apm.ipfs.rpc "http://localhost:5001#default"
```
I get the same result, and `apmOptions.ipfs.rpc` is still the same as before.

#### Environment

- OS and OS version: PureOS 8

0x6431346e:
# 🦅 Pull Request

- Remove the [hack](https://github.com/aragon/aragon-cli/commit/4bc4f36b083d87490a5bcf4a6931dbe5a8d54122) of overriding `--apm.ipfs.rpc` as now we have the better fix: apm options in environments, this should close #514 and close #445 
- Add the `strict` option to yargs (it is easy to make a typo like `enviromnent` and not understand what's wrong)
- Removed a bunch of `parserConfiguration` options that were the same as the default
- Fixed an issue introduced when we upgraded to yargs v13 causing `--silent` and `--debug` to not work anymore (the type was not boolean, so when passing `--debug` and not `--debug=true`, the option was considered undefined and defaulted to false)
- `dao apps` will not show `kernel@vundefined` anymore (although I'm not sure why is not finding the version to begin with)
- Pins `@aragon/wrapper` to `v5.0.0-rc.9` to fix, issue related https://github.com/aragon/aragon.js/issues/325:
```
→ Could not create transaction due to missing app artifact
```

- Update client to 0.7.4

## ✔️ PR Todo


packages/aragon-cli/src/cli.js
```diff
@@ -109,7 +109,11 @@ cmd.option('apm.ipfs.rpc', {
   description: 'An URI to the IPFS node used to publish files',
   default: 'http://localhost:5001#default',
 })
-cmd.group('apm.ipfs.rpc', 'APM providers:')
+cmd.option('apm.ipfs.gateway', {
+  description: 'An URI to the IPFS Gateway to read files from',
+  default: 'http://localhost:8080/ipfs',
```
sohkai: Wondering if we should default this to `ipfs.eth.aragon.network/ipfs` instead, since most users who aren't running this in an aragon app directory (e.g. just using the `dao` commands to configure a dao) won't have an environment applied.

PR #545, opened 2019-06-21
Fix e2e test setup

0xGabi:
# 🦅 Pull Request

Update to fix the problem of two test trying to spawn `aragon devchain`


packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js
```diff
@@ -129,14 +129,13 @@ function sanityCheck(
   modulePath,
   artifact
 ) {
-  const { environments, contractPath } = artifact
-  const { appName, registry, network } = environments[0]
```
0x6431346e: It was failing here because `environments` is not an array.
That being said `environments['default']` does not feel alright either.

packages/e2e-tests/src/cli/apm-publish.test.js
```diff
@@ -1,45 +1,22 @@
...
-test.beforeEach(() => {
-  fs.ensureDirSync(testSandbox)
-})
-
-test.afterEach(() => {
-  fs.removeSync(testSandbox)
-})
-
-test('should publish an aragon app directory successfully', async t => {
+test.skip('should publish an aragon app directory successfully', async t => {
```
0x6431346e: I've marked it to be skipped because, even though it passes, it makes `aragon run` fail: `Cannot find transaction pathing`.

Not sure if it's possible to resolve this conflict, but perhaps we shouldn't test deploying to the local devchain since `aragon run` is already doing it. Can we try deploying to rinkeby instead?

0xGabi: I tried to use Rinkeby but I'm also having problems with `Cannot find transaction pathing` on some node versions. I'll keep the `.skip` and merge it

0x6431346e:
LGTM!

packages/e2e-tests/src/cli/apm-publish.test.js
0x6431346e: Oh, I see, it's the same issue as with local host:
It cannot find the transaction path when another deploy is happening at the same time.

Any thoughts @sohkai 	?

PR #555, opened 2019-06-25
Update dao token new options

### LINKED ISSUE #561: Do not deploy a minime factory for each `dao token new`
## Reduce gas cost for deploying a new token

Currently `dao token new` deploys a minime token factory, then deploys a new token using that factory. This is quite wasteful, the point of having a factory is to re-use it. 🙃

## Have you read the [Contributing Guidelines on issues](https://github.com/aragon/aragon-cli/blob/master/CONTRIBUTING.md#ways-to-contribute)?

No

## Motivation

The cost to deploy a token is toodamnhigh.jpg

## Pitch

This should significantly reduce the cost to deploy a new token from the CLI.

0xGabi:
# 🦅 Pull Request

close #561 

Fix option `transfer-enable` (https://github.com/aragon/aragon-apps/issues/630#issuecomment-505272954)

Add new option `token-factory-address` allowing to provide an address of a deployed MiniMeTokenFactory instead of deploying a new one every time. It also include both Rinkeby and Mainnet default factories if one is not provided.

## ✔️ PR Todo


0x6431346e:
Pushed a commit to add parsing for `fnArgs`

```sh
daniel@zen5-ub:~$ dao exec adasd asdas asd 1 false "true" '["0xad","false"]' '0' '{"foo":"bar"}' --debug
fn-args before parsing [ '1', 'false', 'true', '["0xad",1,"false"]', '0', '{"foo":"bar"}' ]
fn-args after parsing  [ 1, false, true, [ '0xad', 1, 'false' ], 0, '{"foo":"bar"}' ]
```

Note: It will not parse objects (I suspect there isn't a scenario where you want to send an object)

0xGabi:
I test it with initialize of the Token Manager and the problem is still happening 🤕

packages/aragon-cli/src/commands/dao_cmds/token_cmds/new.js
```diff
@@ -19,23 +21,28 @@ exports.builder = yargs => {
...
       description: 'Whether the new token will have transfers enabled',
+      boolean: true,
       default: true,
     })
-    .option('decimal-units', {
-      description: 'Total decimal units the new token will use',
-      default: 18,
+    .option('token-factory-address', {
+      description: 'Address of the MiniMeTokenFactory',
+      type: 'string',
```
sohkai: Given that we know at least one of these factories exists in both rinkeby and mainnet, it would be nice to default to them if the user's environment matches those networks.

0xGabi: I searched on the deployments repo and couldn't found these factories. Are A1 using some deployed by Giveth or someone else?

0xGabi:
I will merge this as it is and we can keep working to fix the issue with Token Manager initialize function on a new PR

packages/aragon-cli/src/commands/dao_cmds/token_cmds/new.js
izqui: For mainnet, this is the `MiniMeTokenFactory` that all the kits use: https://etherscan.io/address/0x909d05f384d0663ed4be59863815ab43b4f347ec (it's linked in the [deployment documentation](https://github.com/aragon/deployments/tree/master/environments/mainnet#dao-kits-1) for mainnet)

For rinkeby, we don't use a canonical factory as the kit deployment script deploys a new one every time. This is the `MiniMeTokenFactory` that the latest version of `democracy-kit` uses: https://rinkeby.etherscan.io/address/0x6ffeb4038f7f077c4d20eaf1706980caec31e2bf

sohkai: Actually, it may make sense to deploy a new instance of these factories to both rinkeby and mainnet, so we can see how users interact with this command (at least with the default factory specified).

PR #570, opened 2019-06-30
Refactor artifact generation & improve sanity check

### LINKED ISSUE #365: apm publish should generate code.sol on minor and patch versions as well
Right now, `code.sol` is only generated on major version changes (i.e. when the contract actually changes).

It would be easier for frontends if each version, including minor and patch versions, had this file so they didn't have to look at the last major version's content to find it.

0xGabi:
# 🦅 Pull Request

close #365

Improve `apm publish` artifact generation task. 
Generate `code.sol` for minor and patch.

## ✔️ PR Todo

### DIFF
--- a/packages/aragon-cli/src/commands/apm_cmds/publish.js
+++ b/packages/aragon-cli/src/commands/apm_cmds/publish.js
@@ -27,6 +27,7 @@ const {
   getMajor,
   sanityCheck,
   generateApplicationArtifact,
+  generateFlattenedCode,
   copyCurrentApplicationArtifacts,
 } = require('./util/generate-artifact')
 
@@ -341,14 +342,18 @@ exports.task = function({
         title: 'Generate application artifact',
         skip: () => onlyContent && !module.path,
         task: async (ctx, task) => {
-          async function invokeAction(answer) {
+          async function invokeArtifactGeneration(answer) {
             if (POSITIVE_ANSWERS.indexOf(answer) > -1) {
               await generateApplicationArtifact(
                 cwd,
+                apm,
                 dir,
                 module,
-                ctx.deployArtifacts
+                ctx.deployArtifacts,
+                web3,
+                reporter
               )
+              await generateFlattenedCode(dir, module.path)
               return `Saved artifact in ${dir}/${ARTIFACT_FILE}`
             }
             throw new Error('Aborting publication...')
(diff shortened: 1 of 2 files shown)


packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -341,93 +343,98 @@ exports.task = function({
...
                 {
                   validate: value => {
                     return ANSWERS.indexOf(value) > -1
                   },
-                  done: async answer => invokeAction(answer),
+                  done: async answer => invokeArtifactGeneration(answer),
                 }
               )
+            } else {
+              return task.skip('Using existent artifact')
```
sohkai:
```suggestion
              return task.skip('Using existing artifact')
```

0xGabi:
Yes `artifact.json#environments` only include the environment been deployed. 

This was how the artifacts were generated so far. I haven't change that but I did include a sanity check to regenerate the artifacts as you mention. 

Do you think we should include all environments instead? And might be the case that this is an old behavior from using `network` option?

0x6431346e:
> Do you think we should include all environments instead? And might be the case that this is an old behavior from using `network` option?

I want to say yes, but I don't know where and how they are use. 

Would be much nicer UX if let's say the `token-manager@7.0.0` had the same contentURI on mainnet and rinkeby. :thinking: 	

What do you think @sohkai 	?

sohkai:
I actually do like that the published artifact only has the correct environment specified, since it's then easier to know what was published... but having the same IPFS hash is also very useful 😄.

I think this is the only bit that would change the IPFS hash between environments, so I am now leaning towards keeping all the environments for the sake of simplicity (especially since the artifact file is the one that holds the ABI, and resolving it is key to doing everything).

packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js
```diff
@@ -47,18 +58,16 @@ async function deprecatedFunctions(apm, artifact, web3, reporter) {
...
           }
         } else {
-          // TODO: (Gabi) Handle warning messages
-          // reporter.warning(
-          //   `Cannot find artifacts for version ${version.version} in aragonPM repo. Please make sure the package was published and your IPFS or  HTTP server are running.`
-          // )
-          return deprecated
+          reporter.warning(
+            `Cannot find artifacts for version ${version.version} in aragonPM repo. Please make sure the package was published and your IPFS or HTTP server are running.`
+          )
```
sohkai: We may want to also ask the user if they'd like to abort here... if I wanted deprecated methods but the old CID wasn't fetchable I'd want to stop the publishing process.

0x6431346e:
> I think this is the only bit that would change the IPFS hash between environments, so I am now leaning towards keeping all the environments for the sake of simplicity (especially since the artifact file is the one that holds the ABI, and resolving it is key to doing everything).

Would it make sense to include all environments in `artifact.json` and then extract the current one to `environment.json` ?

Also, is it used anywhere currently? Does `aragon.js` actually use the `ensRegistry` specified in `artifact.json` ?

sohkai:
AFAIK it is not used anywhere.

Extracting it into an `environment.json` would still change the root dir's CID, but would keep the artifact.json's CID intact? For simplicity, it might make sense just to keep the entire root CID the same across environments :shrug: (apps are usually built to be network-generic) :shrug:.

0xGabi:
Great discussion, I'll create a new issue to update `artifact.json` to use all environments on other PR.

0x6431346e:
> Extracting it into an `environment.json` would still change the root dir's CID, but would keep the artifact.json's CID intact? 

Yes that would be an improvement.

> For simplicity, it might make sense just to keep the entire root CID the same across environments :shrug: (apps are usually built to be network-generic) :shrug:.

Totally agree! especially since this is not used anywhere. It feels like this metadata should belong to APM.


1. I imagine it would make generating and maintaining environments much easier: if I want to deploy `voting.aragonpm.eth` to xDAI all I have to do is a transaction with the `version` and the `contentURI` from mainnet. No need to recompile > less room for errors.
2. Perhaps `aragen` could become a set of scripts that does exactly this: keeps apm registries on par across environments. (We can query the APM repo and download the `ipfs-cache` folder via IPFS).

Issue #369, opened 2019-02-16
`dao exec`, `dao act`: fix support for string, bool and array types in function arguments


izqui:
Currently state:

- `string`: even if using quotes, it will just use the first word and interpret everything after a space as a new argument.
- `bool`: i haven't been able to find a value to send a `false` parameter
- `arrays`: 🤷‍♀️

0x6431346e:
Related to #320

0x6431346e:
> * `string`: even if using quotes, it will just use the first word and interpret everything after a space as a new argument.

Can't figure this out either... but after checking this example:
```
#!/usr/bin/env node

var opts = require('yargs')
    .argv;

console.log(opts);
```
And seeing that either of these work:
```
'hello world'
"hello world"
hello\ world
```

I'm thinking it's something in our configuration/middle-wares :thinking: 	

`bool` and `array` should work once https://github.com/aragon/aragon-cli/pull/555 is merged

0xGabi:
Related to #1008

PR #623, opened 2019-07-12
allow arrays to be passed via --app-init-params

### LINKED ISSUE #320: Improve argument type coercing by inspecting the ABI
When trying to create a new dao from a `multisig-kit`, we need to call `newTokenAndInstance` with 
```
       string name,
       string symbol,
       address[] signers,
       uint256 neededSignatures
```
Note that signers here needs to be an array, so this won't work:

```
> aragon dao new multisig-kit --fn newTokenAndInstance --fn-args Aragon ANT 0xb4124cEB3451635DAcedd11767f004d8a28c6eE7 1

expected array value (arg="signers", coderType="array", value="0xb4124cEB3451635DAcedd11767f004d8a28c6eE7")
```

This won't work either:
```
> aragon dao new multisig-kit --fn newTokenAndInstance --fn-args Aragon ANT '["0xb4124cEB3451635DAcedd11767f004d8a28c6eE7"]' 1
```
because signers is still a string:
```
fnArgs:  [ 
  'Aragon',
  'ANT',
  "['0xb4124cEB3451635DAcedd11767f004d8a28c6eE7']",
  '1' 
]
```

We could do some parsing similar to [yargs/docs/tricks](https://github.com/yargs/yargs/blob/master/docs/tricks.md), i.e.: 
- if `JSON.parse(arg)` works then override with that

Thoughts?

mikec:
web3.eth.abi.encodeFunctionCall expects a javascript array, which needs to be parse from the CLI input

# 🦅 Pull Request

Closes #320

run command doesn't support arrays as --app-init-params

## 🚨 Test instructions

create a contract that takes an `address[]` type as a param for the `initialize()` function. Try executing run command with `--app-init-params ["<address>"]`

### DIFF
--- a/packages/aragon-cli/src/commands/dao_cmds/utils/encodeInitPayload.js
+++ b/packages/aragon-cli/src/commands/dao_cmds/utils/encodeInitPayload.js
@@ -5,6 +5,14 @@ module.exports = (web3, abi, initFunctionName, initArgs) => {
     return '0x'
   } else {
     try {
+      // parse array parameters from string inputs
+      for (var i in methodABI.inputs) {
+        if (methodABI.inputs[i].type.includes('[')) {
+          initArgs[i] = JSON.parse(
+            initArgs[i].replace(new RegExp("'", 'g'), '"')
+          )
+        }
+      }
       return web3.eth.abi.encodeFunctionCall(methodABI, initArgs)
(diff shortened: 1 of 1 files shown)


0x6431346e:
Solves #320 
Related to #369

PR #635, opened 2019-07-18
Project name validation

### LINKED ISSUE #147: Illegal char error when using illegal ENS chars in app name during aragon init
Getting this after running `yarn start` in a project bootstrapped with `aragon init`

```
~/work/src/github.com/althea-mesh/try_aragon/my_app_name34
$ yarn start
yarn start v0.24.6
$ aragon run 
 ✔ Start a local Ethereum network
 ✔ Check IPFS
 ❯ Publish app to APM
   ✔ Deploy contract
   ↓ Automatically bump version [skipped]
     → Starting from initial version
   ✔ Determine contract address for version
   ✔ Building frontend
   ✔ Prepare files for publishing
   ✖ Generate application artifact
     → Illegal char _
     Publish my_app_name34.aragonpm.eth
     Fetch published repo
   Create DAO
   Open DAO
 ✖ Illegal char _

```

macor161:
# 🦅 Pull Request

Closes #147. Displays an error during initialization when project name contains invalid chars.

## 🚨 Test instructions

`aragon init my_app_name`
or
`create-aragon-app my_app_name`

## ✔️ PR Todo

- [ ] (N/A) Update the relevant documentation

### DIFF
--- a/packages/aragon-cli/src/util.js
+++ b/packages/aragon-cli/src/util.js
@@ -302,6 +302,15 @@ const parseStringIfPossible = target => {
   return target
 }
 
+/**
+ * Validates an Aragon ENS subdomain name
+ * @param {string} name Name
+ * @returns {boolean} `true` if valid
+ */
+function isValidEnsName(name) {
+  return /^[a-z0-9]+$/.test(name)
+}
+
 module.exports = {
   parseStringIfPossible,
   debugLogger,
@@ -314,6 +323,7 @@ module.exports = {
   getLocalBinary,
   getGlobalBinary,
   getContract,
+  isValidEnsName,
   ANY_ENTITY,
   NO_MANAGER,
   ZERO_ADDRESS,
--- a/packages/aragon-cli/src/util.js
+++ b/packages/aragon-cli/src/util.js
@@ -308,7 +308,7 @@ const parseStringIfPossible = target => {
  * @returns {boolean} `true` if valid
  */
 function isValidEnsName(name) {
-  return /^[a-z0-9]+$/.test(name)
+  return /^[a-z0-9-]+$/.test(name)
 }
 
 module.exports = {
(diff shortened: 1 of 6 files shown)


macor161:
hm.. seems like some CI checks are failing. I'm looking into this.

packages/aragon-cli/src/util.js
```diff
@@ -302,6 +302,15 @@ const parseStringIfPossible = target => {
   return target
 }
 
+/**
+ * Validates an Aragon ENS subdomain name
+ * @param {string} name Name
+ * @returns {boolean} `true` if valid
+ */
+function isValidEnsName(name) {
+  return /^[a-z0-9]+$/.test(name)
```
0xGabi: @macor161 ENS allow the use of the `-` symbol as well.

macor161: Yes I was wondering whether to add the hyphen or not, since it is not accepted in the main onboarding app (app.aragon.org). Is there an official documentation where I can find the accepted chars for Aragon subdomains?

0xGabi: You are right about that. Nevertheless in that cause those are to link a dao to an [`aragonId`](https://github.com/aragon/aragon-id). We are creating a subdomain in the `open.aragonpm.eth` registrar for Aragon apps. You think we should also stick to not use the hyphen? 

Sorry I went ahead and made the changes 😅

sohkai: The choice to not use the hyphen is a purely arbitrary decision made by us, and not an ENS limitation. You can use any [IRI](https://www.ietf.org/rfc/rfc3987.txt) (including emojis, etc.).

The frontend actually accepts [this regex](https://github.com/aragon/aragon/blob/master/src/web3-utils.js#L170) for self-assigned ENS names. Note that again, it is still a subset of possible ENS names, but it was chosen to avoid potential spoofing attacks on the frontend with unicode characters.

macor161: @sohkai Thanks for clarifications! 
@0xGabi Thanks for those light-speed fixes :) 
So after discussing the subject, we decided to keep the hyphen since it was already possible to create apps that contain them with the current CLI.

PR #646, opened 2019-07-19
Use all environment on artifact generation

### LINKED ISSUE #585: Update artifact.json generation to use all evnrionments
## Motivation
Right now we only include the environment been publish to on the `artifact.json` this cause that the IPFS cid is different across environments.

Further discussion: https://github.com/aragon/aragon-cli/pull/570#issuecomment-508557876

0xGabi:
# 🦅 Pull Request

closes #585


sohkai:
Do we have a way on `apm publish` to specify a CID and skip the build process?

0xGabi:
No we don't. Seems quite useful will add it in a different PR 👍

PR #660, opened 2019-07-24
Fix prepare files function to handle --files correctly

### LINKED ISSUE #581: Arguments to `apm publish --files` are filtered out by the ignore list
#### Environment

- `aragonCLI 5.6.2` (but think this applies to all future versions too)

## To Reproduce

1. Clone https://github.com/aragonone/approvals/tree/d4430154936f1f87ecf123cd449684c5736267c2
1. After doing `npm install` in the root and in the `app` folder, run `npm start` in the root
1. The Aragon client starts, but when you try to go to the Approvals app, it'll say it can't resolve an `index.html` file

This is because the project has a `.gitignore` file, that includes `build`. We manually specify `--files app/build` during the publish (and run) process, and despite specifying it, the gitignore pattern gets applied and ignores this.

## Expected behavior

Files specified through `--files` would override the ignore list.

## Actual Behavior

Files specified through `--files` are ignored by the `.gitignore` or `.ipfsignore`.

Note that we were able to fix this by adding a blank `.ipfsignore` file (see [commit](https://github.com/aragonone/approvals/commit/4e7527e3c14afcb3645a2248ab75c442042c1e92)), which is OK, but a bit confusing :).

0xGabi:
# 🦅 Pull Request

closes #581


packages/aragon-cli/scripts/extract-roles.js
```diff
@@ -20,11 +20,16 @@ const aOSRoles = [
   { id: 'APP_MANAGER_ROLE', name: 'Manage DAO apps' },
 ]
```
sohkai: Maybe we should split this list into `kernelRoles` and `aclRoles` too (`CREATE_PERMISSIONS_ROLE` is for the ACL, `APP_MANAGER_ROLE` is for the kernel)

packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js
```diff
@@ -47,7 +47,7 @@ function decorateFunctionsWithAbi(functions, abi, web3) {
 }
 
 async function deprecatedFunctions(apm, artifact, web3, reporter) {
-  let deprecated = {}
+  let deprecatedFunctions = {}
```
sohkai: I think this can be declared `const`

packages/aragon-cli/src/commands/apm_cmds/util/preprare-files.js
```diff
@@ -25,21 +25,15 @@ async function prepareFilesForPublishing(
...
+    const ipfsignoreFile = fs.readFileSync(ipfsignorePath).toString()
+    filter.add(ipfsignoreFile)
   } else {
     const gitignorePath = path.resolve(projectRoot, '.gitignore')
     if (pathExistsSync(gitignorePath)) {
-      filter.add(createFilter(files, gitignorePath))
+      const gitignoreFile = fs.readFileSync(gitignorePath).toString()
+      filter.add(gitignoreFile)
     }
   }
```
sohkai: It seems to work fine without, but [`ignore`'s pathname conventions](https://www.npmjs.com/package/ignore#pathname-conventions) says we should strip out the `.` and use `path.join()` in `filterIgnoredFiles()` 🤷‍♂

packages/aragon-cli/config/truffle.default.js
```diff
@@ -10,7 +10,10 @@ const HDWalletProviderPrivkey = require('truffle-hdwallet-provider-privkey')
 const DEFAULT_MNEMONIC =
   'explain tackle mirror kit van hammer degree position ginger unfair soup bonus'
 
-const defaultRPC = network => `https://${network}.infura.io`
+const ARAGONCLI_PROJECT_ID = 'd146fdf20d88476d832edd5d65fc2339'
+
+const defaultRPC = network =>
+  `https://${network}.infura.io/v3/${ARAGONCLI_PROJECT_ID}`
```
sohkai: Hmm, don't know about this; is `mainnet.eth.aragon.network` too slow?

0xGabi: It was cause we had a [`Legacy access request rate exceeded` issue](https://community.infura.io/t/legacy-access-request-rate-exceeded/490) when doing something like: `aragon apm packages open.aragonpm.eth`. I think cause there are too many repos on that one.

0xGabi: I create an Infura Project id with my account, but maybe Aragon already have one?

sohkai: We do have one, but we shouldn't embed it into the code like this (because people may find it and abuse it).

Why not just use `https://mainnet.eth.aragon.network`?

0xGabi: Oh cool, that works as well. This was a legacy value I was reusing, but definitely we should use Aragon's rpc 👍

PR #678, opened 2019-08-01
New command devchain status

0xGabi:
# 🦅 Pull Request


packages/aragon-cli/src/commands/dao_cmds/token_cmds/new.js
```diff
@@ -11,7 +11,7 @@ const {
 } = require('../../../util')
 
 const MAINNET_MINIME_TOKEN_FACTORY =
-  '0x909d05f384d0663ed4be59863815ab43b4f347ec'
+  '0xA29EF584c389c67178aE9152aC9C543f9156E2B3'
```
sohkai: @0xGabi Is this a newly deployed version of the minime factory? We should verify its code on etherscan.

PR #687, opened 2019-08-06
CLI: avoid coercing number arguments from strings to JS numbers

sohkai:
# 🦅 Pull Request

```
dao act 0x510846E270759CC92131D3453c58e1073C191cD1 0xec67005c4e498ec7f55e092bd1d35cbc47c91892 "transfer(address,uint256)" 0x0d580ae50B58fe08514dEAB4e38c0DFdB0D30adC 1000000000000000000 --environment aragon:mainnet
```

Results in an

```
invalid number value (arg="", coderType="uint256", value=1000000000000000000)`
```

Error produced from web3-eth-abi.

## 🚨 Test instructions

Run

```
dao act 0x510846E270759CC92131D3453c58e1073C191cD1 0xec67005c4e498ec7f55e092bd1d35cbc47c91892 "transfer(address,uint256)" 0x0d580ae50B58fe08514dEAB4e38c0DFdB0D30adC 1000000000000000000 --environment aragon:mainnet
```

## ✔️ PR Todo

--------------

Note that web3-eth-abi does not like much else than a string; giving it a BN.js object or etc. does not work :(.

In general though, unless we need to do computation to a number, it's often best to keep it around as a string since Ethereum numbers are generally _large_.

### DIFF
--- a/packages/aragon-cli/src/commands/dao_cmds/act.js
+++ b/packages/aragon-cli/src/commands/dao_cmds/act.js
@@ -2,7 +2,7 @@ const ABI = require('web3-eth-abi')
 const execHandler = require('./utils/execHandler').handler
 const getAppKernel = require('./utils/app-kernel')
 const { ensureWeb3 } = require('../../helpers/web3-fallback')
-const { parseStringIfPossible, ZERO_ADDRESS } = require('../../util')
+const { parseArgumentStringIfPossible, ZERO_ADDRESS } = require('../../util')
 
 const EXECUTE_FUNCTION_NAME = 'execute'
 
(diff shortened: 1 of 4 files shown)


0x6431346e:
I can see that:
```sh
dao act 0x510846E270759CC92131D3453c58e1073C191cD1 0xec67005c4e498ec7f55e092bd1d35cbc47c91892 "transfer(address,uint256)" 0x0d580ae50B58fe08514dEAB4e38c0DFdB0D30adC 9007199254740991 --environment aragon:mainnet --debug
```
does not throw an error, so perhaps we can do a check:
```js
const number = Number(target)

if (isNaN(number) || number > Number.MAX_SAFE_INTEGER) {
  throw new Error(`Cannot parse ${target} as number`)
}
```

I imagine this utility could be useful in the future (outside the context of web3), but mainly I think parsing only some, and not all data types is not very intuitive. :thinking: 	What do you think?

sohkai:
>  Mainly I think parsing only some, and not all data types is not very intuitive

I agree with this, but only parsing up to the the max safe int is also not very intuitive and you really have no way to know how to handle its output unless you do the same check before, or do a typeof check afterwards.

Perhaps the best strategy would be to parse into a BN.js object, or similar, and write another "coerceIntoWeb3" type of function? That way we separate parsing string arguments from coercing them into web3.

PR #707, opened 2019-08-12
New Assign Id Command

### LINKED ISSUE #347: Claiming an aragonid
Not sure if this belongs in the CLI, but currently if you create and deploy an organization with the CLI you are not able to claim an aragonid as you would if creating an organization using the standard onboarding process. 

It would be nice if organization created using the CLI could be accessed in the same way (eg  mainnet.aragon.org/#/[id].aragonid.eth

macor161:
# 🦅 Pull Request

Closes #347

Adds: 
- A `dao assign-id <dao> <aragon-id>` command 
- An `--aragon-id <aragon-id>` param for the `dao new` command

I’m not entirely sure of the command name `dao assign-id`. It is not consistent with other commands as it contains a hyphen. However, I feel `dao assign` may be a bit too vague. What do you think?

## ✔️ PR Todo


packages/aragon-cli/src/commands/dao_cmds/assign-id.js
```diff
@@ -0,0 +1,127 @@
...
+              process.emit(1)
+            }
+          } catch (err) {
+            // ens.resolver() throws an ENS.NameNotFound error if name doesn't exist
+            if (err !== ENS.NameNotFound) throw err
+          }
+        },
+      },
+      {
+        title: 'Assigning Id',
```
0xGabi: I notice this task is executed even when id can't be assigned. Maybe we can pass `exists` throught `ctx` and only enable the task if `!ctx.exists`?

packages/aragon-cli/src/commands/dao_cmds/assign-id.js
```diff
@@ -0,0 +1,127 @@
+const TaskList = require('listr')
+const ENS = require('ethereum-ens')
+const { ensureWeb3 } = require('../../helpers/web3-fallback')
+const { green } = require('chalk')
+const listrOpts = require('@aragon/cli-utils/src/helpers/listr-options')
+const { isAddress } = require('web3-utils')
+const registrarAbi = require('@aragon/id/abi/IFIFSResolvingRegistrar').abi
+const { sha3 } = require('web3-utils')
+
+const ARAGON_DOMAIN = 'aragonid.eth'
```
0xGabi: @0x6431346e we should remember to update to v7

macor161:
> So we have `dao id assign` (that could be an alias of `dao id`) and eventually we might include `dao id transfer`.

I like that! It would also leave room to eventually add an `unassign` command. I will post an update today.

packages/aragon-cli/src/commands/dao_cmds/assign-id.js
macor161: Nice catch! The problem is caused by a silly typo. It should be `process.exit(1)` instead of `process.emit(1)` on line `72`.

sohkai:
Haha while `unassign` would be nice, unfortunately there's no easy way to accomplish that (yet). We'd have to update the Kernels to interact with ENS to do this, and it's not a change I'm particularly keen on doing soon.

macor161:
Ah good to know! Thanks Brett!

macor161:
Just updated the command name to `dao id assign <dao> <aragon-id>` with a shortcut `dao id <dao> <aragon-id>`

PR #709, opened 2019-08-12
v7: Extensions refactor

0x6431346e:
# 🦅 Pull Request

## ✔️ PR Todo

- [ ] Include links to related issues/PRs


packages/cli-utils/src/middleware/environment.js
```diff
@@ -0,0 +1,187 @@
+const Web3 = require('web3')
```
0xGabi: What do you think of using a similar `environment.js` file as: https://github.com/aragon/aragon-cli/blob/91200d3f2f0051b45a344d04b555e242ad666041/packages/aragon-cli/src/middleware/environment.js

0xGabi:
@0x6431346e Can you instead do it from https://github.com/aragon/aragon-cli/pull/797 I think I have included a couple of additional clean ups to the code.

PR #734, opened 2019-08-22
Check IPFS & install if missing

### LINKED ISSUE #716: Minor, out-of-the box bugs with create-aragon-app
## 🐛 Bug Report

The following 3 bugs occur after a clean run of `npx create-aragon-app test`:

1) Running create-aragon-app, then `npm start` for the first time, throws an ipfs error with the message 
"IPFS not installed. Use `aragon ipfs install` before proceeding". This command fails if you run it, since it uses a globally installed aragon cli. If you use `npx aragon ipfs install` instead, it works.

2) Running `npm test` fails. The script ganache-cli uses `npx truffle test`, but there is no Truffle dependency installed in the boilerplate React project, so it assumes that you have a global truffle version installed. If you don't, it'll use npx and hence lock into the latest version, Truffle 5, which fails because the project needs to use Truffle 4. The boilerplate React project should have the Truffle 4 dependency installed locally.

3) This one is super minor, but CounterApp tests are missing an `await` in app.initialize(). No apparent effects, but the calls are async.

0xGabi:
# 🦅 Pull Request

closes #716


ajsantander:
@0xGabi what about the error message that prompted you to install ipfs. Should we delete that prompt as well? If we're not removing it, should we modify the message so that it uses npx?

0xGabi:
Actually there is two scenarios:
1. Using `create-aragon-app` we should just check if the user have IPFS installed and if not we just install it
2. Using `aragonCLI` in this case we display the error when use need to install IPFS, then they are able to use `aragon IPFS install` without npx cause they have it installed globaly

PR #767, opened 2019-09-11
Added CLI Mnenmonic Warning

### LINKED ISSUE #758: Initialize every Aragon CLI download with it's own seed phrase
## 💥 Proposal

AFAIK the Aragon CLI uses the same accounts from the same seed phrase for everyone. That's been fine mostly, but today I just tried to deploy a Rinkeby or and realized that the account was out of Rinkeby ETH. My first thought was, whatever I'll just add more. Then it dawned on me that actually anyone could take ownership of my DAO in the middle of my deployment. Even though it's just a Rinkeby test, this is still not good. It would be great if every download of the Aragon CLI came with a unique seed phrase.

burrrata:
Added a placeholder warning regarding the Aragon CLI default mnemonic. Not sure if there's a guide on how to use your own signing key for the CLI (I think I saw one somewhere).

Does not close #758, but hopefully it helps.

### DIFF
--- a/docs/Dao-commands.md
+++ b/docs/Dao-commands.md
@@ -1,5 +1,8 @@
 The `aragon dao` commands can be used for interacting with your DAO directly from the command line. These commands are also available directly using the `dao` alias.
 
+> **Note**<br>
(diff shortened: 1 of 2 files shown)


0xGabi:
Thanks @burrrata. The the link you are looking for is: https://hack.aragon.org/docs/guides-faq#set-a-private-key

burrrata:
@0xGabi Thanks! Added the link and also copied the warning to the DAO commands page where people would be using the CLI to create DAOs

0xGabi:
This is much clear now! Thanks

PR #784, opened 2019-09-20
Add deploy event compatibility

macor161:
# 🦅 Pull Request

Some templates use `DeployDAO` events instead of `DeployDao`. This PR adds support for those events.

## ✔️ PR Todo

### DIFF
--- a/packages/aragon-cli/src/commands/dao_cmds/new.js
+++ b/packages/aragon-cli/src/commands/dao_cmds/new.js
@@ -137,7 +137,9 @@ exports.task = async ({
           // Backward compatibility with old event name
           const deployEventValue =
             events[deployEvent] ||
-            events[exports.OLD_BARE_TEMPLATE_DEPLOY_EVENT]
+            events[exports.OLD_BARE_TEMPLATE_DEPLOY_EVENT] ||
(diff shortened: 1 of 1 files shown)


packages/aragon-cli/src/commands/dao_cmds/new.js
```diff
@@ -137,7 +137,9 @@ exports.task = async ({
           // Backward compatibility with old event name
           const deployEventValue =
             events[deployEvent] ||
-            events[exports.OLD_BARE_TEMPLATE_DEPLOY_EVENT]
+            events[exports.OLD_BARE_TEMPLATE_DEPLOY_EVENT] ||
+            // Some templates use DeployDAO instead of DeployDao
+            events['DeployDAO']
```
sohkai: 🤦‍♂ are these the old kits?

macor161: The problem was in the boilerplate: https://github.com/aragon/aragon-react-boilerplate/blob/558074e1863bbcb44cb0d2c6470866bbd94ca40e/contracts/Template.sol#L30
But it seems to be fixed now.

PR #796, opened 2019-09-25
Update wrapper to rc.17

macor161:
# 🦅 Pull Request

Update wrapper to `rc.17`.
It seems that when `wrapper.getApp(...)` is called before the app is loaded, it returns `undefined`. So I added a check before calling `getTransactionPath`.

## 🚨 Test instructions

`aragon apm publish major`

## ✔️ PR Todo


packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -545,7 +547,20 @@ exports.runPublishTask = ({
         enabled: () => !onlyArtifacts,
         task: async (ctx, task) => {
           try {
-            const getTransactionPath = wrapper => {
+            const getTransactionPath = async wrapper => {
+              // Wait for app info to load
+              await wrapper.apps
```
sohkai: > It seems that when `wrapper.getApp(...)` is called before the app is loaded, it returns undefined. So I added a check before calling `getTransactionPath`.

This is indeed the case; none of the wrapper's functionality waits for it to finish loading the apps as it assumes the client will be reactive based on its streams.

While this should be made moot in other ways, e.g. a different, lighter, non-reactive JS-based API, for now we may want to encapsulate this logic into a single place. We may also want to give the user some feedback that we're waiting for their apps to load. Note that this also hangs if the user input is a non-installed app.

macor161: > for now we may want to encapsulate this logic into a single place

Yeah I agree, not very DRY at the moment. I will try to refactor this next week.

> Note that this also hangs if the user input is a non-installed app.

Good catch! Will also try to fix this during the refactoring.

sohkai: > Note that this also hangs if the user input is a non-installed app.

We don't pass any sync information back from the wrapper about the permissions (we certainly could, especially as it starts taking longer and longer to load old orgs), but for now if you've received apps from the observable but the declared app is not in the list, you can assume that it's not installed.

PR #807, opened 2019-10-04
Support ACL permissions with parameters

### LINKED ISSUE #407: Support permissions with parameters
This issue was funded with **240 DAI**, see more issues on our dao [projects app](https://rinkeby.aragon.org/#/meshteam.aragonid.eth/0x2b2290c2370cbc59e7c77bd36072f801d5e996c8).
---

macor161:
# 🦅 Pull Request

Add support for ACL parameters:
```bash
dao acl grant <dao-addr> <app-proxy-addr> <role> <entity> [params...]
```
Each parameter must have the following syntax: `"<id>,<op>,<value>"`. Multiple parameters must be separated by spaces.

See the [ACL documentation](https://hack.aragon.org/docs/aragonos-ref#parameter-interpretation) for the list of available operators.

Closes #407 

## 🚨 Test instructions

```bash
dao acl grant <dao> <app> <role> <entity> "0,GT,2"
```

```bash
dao acl grant <dao> <app> <role> <entity> "2,EQ,0xC7f8dDbc7B3BFd432dEAc0CA270110467EcE01c3"
```

```bash
dao acl grant <dao> <app> <role> <entity> "LOGIC_OP_PARAM_ID,OR,(1,2)"  "0,LT,4"  "1,EQ,42"
```

```bash
dao acl grant <dao> <app> <role> <entity> "LOGIC_OP_PARAM_ID,IF_ELSE,(1,2,3)" 
 "BLOCK_NUMBER_PARAM_ID,GT,1000"  "0,EQ,42"  "0,EQ,0"
```

## ✔️ PR Todo


packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/params.js
```diff
@@ -0,0 +1,156 @@
+const { isString } = require('lodash')
```
sohkai: Someone get the fire trucks, cause this is hot 🔥:fire: :fire:

Haven't checked the correctness (we should certainly add more tests), but:

- We should scope this out into its own package
- We may be able to define a light DSL language for this and use something like [nearley](https://nearley.js.org) or [peg.js](https://pegjs.org) to parse it; we would most likely want to use this in a browser environment so we'll have to be conscious of the parser's size (and obviously handrolling it like now would be the most optimal)
- We should add a decoder to go from the param bytes, for use in UI

macor161: Hahaha thanks! :smile: 
Definitely good ideas. I added them to my to-do list. I'm not super familiar with DSLs howver, I will have to read the subject a little bit more. But point 1 and 3 should be fairly straightforward to do.

PR #816, opened 2019-10-11
Clean dependencies 

0xGabi:
# 🦅 Pull Request

I decided to include this PR to clean a bit before moving to v7 and start the separation of packages.

Update to client `v0.8.2`

@macor161 would you mind take a look?


packages/aragon-cli/package.json
```diff
@@ -67,11 +62,10 @@
     "go-platform": "^1.0.0",
     "http-server": "^0.11.1",
     "ignore": "^4.0.6",
-    "inquirer": "^6.4.1",
+    "inquirer": "^7.0.0",
     "internal-ip": "^4.3.0",
     "ipfs-api": "^26.1.2",
     "ipfs-http-client": "^33.0.1",
-    "js-sha3": "^0.8.0",
```
macor161: I think `js-sha3` is still used at a couple of places.

packages/aragon-cli/src/cli.js
```diff
@@ -35,7 +35,6 @@ cmd.middleware(MIDDLEWARES)
 
 cmd.alias('env', 'environment')
 cmd.alias('h', 'help')
-cmd.alias('v', 'version')
```
macor161: Do we have a good reason to remove this?

0xGabi: Oh this cause what we discuss here: https://github.com/aragon/aragen/pull/79#discussion_r332967208

packages/aragon-cli/package.json
0xGabi: Good catch forgot to update to use `web3.utils`

packages/aragon-cli/src/cli.js
macor161: Ah ok I wasn't aware of the issue between `aragon -v` and `aragon devchain -v`, thanks 🙂 
However `aragon devchain -v` doesn't seem to be working either, is this normal?

0xGabi: Oh sorry about that, it was cause I needed to update aragen first. I did that yesterday

PR #833, opened 2019-10-23
Import ABIs from `@aragon/os` and `@aragon/id`, to resolve #818

### LINKED ISSUE #818: Import ABIs from `@aragon/os` and `@aragon/id` instead of hardcoding them
Rather than bundling their ABIs directly, these are exposed from the npm packages.

dapplion:
# 🦅 Pull Request

- Replace hardcoded ABIs with requires to `@aragon/os` and `@aragon/id`.
- Move the requires from inline to the top of the file alongside other dependencies.
- Standarize naming to `${ContractName}Abi`, in strict camelCase

## 🚨 Test instructions

N/A

## ✔️ PR Todo

- [x] Include links to related issues/PRs. Fixes https://github.com/aragon/aragon-cli/issues/818
- [x] Update unit tests for this change. N/A
- [x] Update the relevant documentation. N/A
- [x] Clear dependencies on other modules that have to be released before merging this. N/A


macor161:
@0xGabi or @0x6431346e can you close this when you have a chance, I can't do it from my account. A review is already in progress for a similar PR: https://github.com/aragon/aragon-cli/pull/837
@dapplion I'm sorry for this, I think that was a really great pull request but we had to make a choice between two identical PRs. Thanks a lot for your contribution, it was very much appreciated.

packages/aragon-cli/package.json
```diff
@@ -45,7 +45,8 @@
     "@aragon/apps-shared-minime": "^1.0.1",
     "@aragon/aragen": "^5.0.0",
     "@aragon/cli-utils": "^0.0.8",
-    "@aragon/truffle-config-v4": "^1.0.0",
```
macor161: We should probably keep `@aragon/truffle-config-v4`

packages/aragon-cli/src/commands/dao_cmds/new.js
```diff
@@ -10,6 +10,7 @@ const {
   getRecommendedGasLimit,
   parseArgumentStringIfPossible,
 } = require('../../util')
+const kernelAbi = require('@aragon/os/build/contracts/Kernel.json').abi
```
macor161: Very small detail but I would remove the `.json` extension to keep consistency with other imports.

PR #835, opened 2019-10-24
Update publish flow to include --dry-run flag

### LINKED ISSUE #760: aragon apm publish --dry-run
## 🚀 Feature

It would great if `aragon apm publish` were able to do a dry run and log the transaction data that needs to be sent to propagate the package publish.

In cases where a deployment **must** be completed without down time or signed later, this feature would allow someone to obtain the necessary transactions ahead of time.

## Motivation

Publishing is tricky and ideally should not lead to downtime if multiple package publishes are necessary. Having all of the transaction data up front would make it easier to ensure that a sequence of publishes will happen regardless of any issues or bugs in the CLI during the publish.

0xGabi:
# 🦅 Pull Request

Close #760

### DIFF
--- a/packages/aragon-cli/src/commands/apm_cmds/publish.js
+++ b/packages/aragon-cli/src/commands/apm_cmds/publish.js
@@ -141,7 +141,7 @@ exports.builder = function(yargs) {
       default: false,
     })
     .option('dry-run', {
-      description: 'Whether to skip the confirmation step',
+      description: 'Whether to broadcast the publish transaction',
       boolean: true,
       default: false,
     })
--- a/packages/aragon-cli/src/commands/apm_cmds/publish.js
+++ b/packages/aragon-cli/src/commands/apm_cmds/publish.js
@@ -140,6 +140,11 @@ exports.builder = function(yargs) {
       boolean: true,
       default: false,
     })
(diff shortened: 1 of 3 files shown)


packages/aragon-cli/src/commands/apm_cmds/publish.js
```diff
@@ -713,9 +723,8 @@ exports.handler = async function({
   )
 
   if (contentProvier === 'ipfs') {
+    reporter.info('Explore IPFS content locally with the following link:')
     console.log(
-      '\n',
-      'Explore the ipfs content locally:',
       '\n',
       chalk.bold(
         `http://localhost:8080/ipfs/QmSDgpiHco5yXdyVTfhKxr3aiJ82ynz8V14QcGKicM3rVh/#/explore/${contentLocation}`
```
sohkai: Having both of these hashes is a little bit confusing; not sure if we can do a great job explaining this, but it would be nice if users could easily understand that this first hash is the IPFS explorer's frontend.

sohkai:
Just tried this, and didn't end up with the tx data, nor the repo version bump:

Screen Shot 2019-10-25 at 1 34 30 AM

PR #837, opened 2019-10-24
Removes hardcoded contract ABIs

### LINKED ISSUE #818: Import ABIs from `@aragon/os` and `@aragon/id` instead of hardcoding them

ajsantander:
Fix #818 

While doing this, I saw that there are a lot of manual `new web3.ethContract(abi, address)` usages. I suggest that we open a new issue to add a "getContract" helper, similar to what already exists in utils.js but capable of retrieving a full Web3 Contract instance.


packages/aragon-cli/package.json
```diff
@@ -45,6 +45,7 @@
     "@aragon/apps-shared-minime": "^1.0.1",
     "@aragon/aragen": "^5.0.0",
     "@aragon/cli-utils": "^0.0.8",
+    "@aragon/id": "^2.1.0",
     "@aragon/truffle-config-v4": "^1.0.0",
```
macor161: We should probably add the `@aragon/os` dependency here

packages/aragon-cli/src/commands/apm_cmds/util/acl.js
```diff
@@ -3,18 +3,27 @@ const { getRecommendedGasLimit } = require('../../../util')
 module.exports = ({ web3, gasPrice, network }) => {
   const getACL = async repoAddr => {
     const repo = new web3.eth.Contract(
-      require('../abi/os/AragonApp').abi,
+      require('@aragon/os/build/contracts/AragonApp').abi,
```
macor161: This syntax definitely works, but I saw a different syntax in other aragon packages. e.g.: https://github.com/aragon/apm.js/blob/master/src/index.js#L159
@sohkai Do you have any recommendation for one or the other?

0xGabi:
@ajsantander this PR have a lot of duplications from #833 

I think we should merge them or decide which one to include.

macor161:
@dapplion `await execa(ipfsBin, ['config', 'show'])` seems to always fail on my laptop (Mac). So it always try to initialize even when it is already initialized. Would you know why?

macor161:
We can close this PR as it will be replaced by #833 
Thanks for your contribution @ajsantander ! 🙌

PR #848, opened 2019-10-28
Sanitize CLI options

### LINKED ISSUE #282: Wrong options provided by `--help` for subcommands
The `aragon devchain --help` command is misleading users by displaying generic options the devchain actually does not take into account: notably `network` and `environment`.

This is due to the fact that these options are defined at the root of the `yargs` entry point and not on a per-command basis:

https://github.com/aragon/aragon-cli/blob/389b08d94b9b0644f7e9a2634180910ee124b8b3/src/cli.js#L85

ajsantander:
WIP

Fix #282

This PR is not just going to fix the issue mentioned, but do a general sanitation of the usage of options in the CLI. There is a lot of duplicate options being declared, so first we need to remove those duplicates. Once that is done, we can begin moving out options out of cli.js and set them specifically in commands.

### DIFF
--- a/packages/aragon-cli/npm-shrinkwrap.json
+++ b/packages/aragon-cli/npm-shrinkwrap.json
@@ -487,6 +487,342 @@
         "listr-silent-renderer": "^1.1.1",
         "listr-update-renderer": "^0.5.0",
         "listr-verbose-renderer": "^0.5.0"
+      },
+      "dependencies": {
+        "ansi-escapes": {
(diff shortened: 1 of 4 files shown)


ajsantander:
I'm closing this PR until after we complete the refactor to separate interface from logic: #849
It doesn't make sense to introduce a big change like this before the refactor.

Nonetheless, this closed PR can be used for future reference.

Issue #851, opened 2019-10-29
Set up branch strategy for SDK v7 refactors


ajsantander:
Since a lot of changes are coming with SDK v7 from the [aragonSDK working group](https://forum.aragon.org/t/aragonsdk-working-group/1411), we need to design and set up a branching and release strategy on the repo, for introducing changes from the refactor while being able to continue delivering stable releases (even though the priority will be the refactors).

The solution should:
* Keep master; release stable versions off of master
* Add develop; bump major version and release nightlies off of develop

PR #863, opened 2019-11-01
Compare contract roles and parameter count with arapp.json, resolves #370

### LINKED ISSUE #370: Analyze role and role params against actual contract usage
The CLI should warn (perhaps at `apm publish`?) if it detects that the roles set in `arapp.json` don't match those in the contract.

dapplion:
# 🦅 Pull Request

Following @sohkai proposal in https://github.com/aragon/aragon-cli/issues/370 this PR adds a "linter" which asserts that the contract roles and arapp.json roles match (same ID and number of params). If not, it will error with:
```
✖ Role usage and declarations in arapp.json do not match: 
 - MANAGE_PAYMENTS_ROLE2 is declared but never used
 - Function setPaymentStatus uses undeclared role MANAGE_PAYMENTS_ROLE
 - Function setPaymentStatus uses EXECUTE_PAYMENTS_ROLE with 2 params but requires 3
```

To do this, a helper traverses the AST instead of using regex like other helpers (`solidity-extractor.js`). I believe this approach is more resilient than regex and more flexible to different code styles. Please note that the `ast-extractor.js` module is fully tested against the Finance and Voting artifacts. All other functions used in this PR are covered by unit tests. Regarding the AST parsing, it does a best effort to resolve expressions in the modifier second parameter but it could be fooled by weird helpers.

**Feedback request**: 
1. Right now the assert is strict error-ing if something is not right. Should it be changed for a non-stopping warning?
2. This functionality is called via `assertContractRoles(cwd, module)` in the `aragon apm publish` command at the beginning of the `runPrepareForPublishTask` and after `runSetupTask` which compiles .sol files, a necessary step for it. Is this the best place to do this check?
3. The tests need build artifacts to run against. Is it ok to have then in the CLI codebase or is there a better approach to get them?

## 🚨 Test instructions

Change the `role`s of an arapp.json so they don't match the ones in the contract. For example, in Finance app change `MANAGE_PAYMENTS_ROLE` name to `MANAGE_PAYMENTS_ROLE2`, and add one extra param to `EXECUTE_PAYMENTS_ROLE`. Then run:
```
lion$ node aragon apm publish major Finance
✖ Role usage and declarations in arapp.json do not match: 
 - MANAGE_PAYMENTS_ROLE2 is declared but never used
 - Function setPaymentStatus uses undeclared role MANAGE_PAYMENTS_ROLE
 - Function setPaymentStatus uses EXECUTE_PAYMENTS_ROLE with 2 params but requires 3
```

## ✔️ PR Todo


0xGabi:
Hey @dapplion wanted you to know that going forward we will follow the brach strategy explained in this issue: https://github.com/aragon/aragon-cli/issues/851

dapplion:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for contributing to Aragon!

Pending until refactor is complete

0xGabi:
We should review and evaluate if we can reuse part of this logic changes on the toolkit.

0xGabi:
This PR is amazing, most of the logic was refactored and improved to use it in the toolkit in #1494

Issue #858, opened 2019-10-31
Refactor entire codebase for separating logic from commands


ajsantander:
## Planning

Planning document, [here.](https://docs.google.com/document/d/14wqx_hEif6M_8g1fqNctdsVSbjVN7InYblr04wZ7CyQ/edit#heading=h.roiaaczav1ux)

Following @0x6431346e's [proposal on how to separate logic from commands](https://github.com/aragon/aragon-cli/pull/856), we're ready to start implementing this in the code.

## Preliminary tasks
- [x] Finish merging outstanding PRs: https://github.com/aragon/aragon-cli/issues/853
- [x] Finish setting up branching strategy for refactors: https://github.com/aragon/aragon-cli/issues/851
- [x] Rebase #856 to master so that it assimilates all the outstanding PR changes
- [x] Merge #856 to the new develop branch

Having completed these preliminary tasks, we're ready to start implementing the refactor, with the following considerations:
* Logic should be extracted to the `lib/` folder following the `token new` command's example (#856).
* Unit tests should be added to all extracted logic, again following the example.
* We shouldn't worry about end-to-end tests for now, but do try to test commands involved with what you're changing before pushing changes.
* Extracted logic should be 100% agnostic, and at most provide updates to the commands using them via callback handlers.
* We can start picking off commands to refactor from the list below, and making independent PRs to the develop branch. Once you pick an issue, add your name next to the item in the list below, or ask one of the repo maintainers to add it for you.

## Refactor based on commands
- [ ] ⬇️ :mountain: **[ Dev: @dapplion , Reviewers: ? ]** start command
- [x] apm commands
  - [x] **[ Dev: @ajsantander, Reviewers: @0xGabi ]** apm info
  - [x] **[ Dev: @ajsantander, Reviewers: @0xGabi, @macor161  ]** apm extract-functions
  - [x] **[ Dev: @ajsantander, Reviewers: @0xGabi, @dapplion  ]** apm versions
  - [x] **[ Dev: @ajsantander, Reviewers: @0xGabi, @macor161  ]** apm packages
- [x] **[ Dev: @dapplion, Reviewers: @0xGabi,  ]** deploy command
- [ ]  ⬇️ :mountain: **[ Dev: @macor161, Reviewers: ? ]** devchain command
- [x]  init command (Deprecated, I'll merge a commit that remove this instead)
- [x] middlewares
[...]

PR #864, opened 2019-11-02
Refactor dao acl - #858 task

dapplion:
# 🦅 Pull Request

- [x] acl create command
- [x] acl grant command
- [x] acl remove-manager command
- [x] acl revoke command
- [x] acl set-manager command
- [x] acl view command

**acl view**:
Split the logic into three files / parts
- `lib/acl/view`: Fetches the data and interfaces with `aragon-js`
```
[ "0xab1200000000000000000000000000000000cd34"  ]
```
- `lib/acl/viewFormatter`: Parses the data and returns the data relevant to the user appended with human names. This component is fully tested
```
{ name: "acl", address: "0xab1200000000000000000000000000000000cd34" }
```
- `commands/dao/acl/view` (original): Call data fetcher, call parser, and then print the data in as a table with emojis on other stringified content.
```
acl (0xab12...cd34)
```
Also, every used function is now typed with JSDoc, and the typings abstracted to `typedef.js`. I have seen that @macor161 also uses JSDoc on new code, can we create general type definitions for reused that structures such a `module`, `network` etc?

**others**:
All other `dao acl` commands just called the wrapper `execHandler` and since they do just one single asynchronous task there is no intermediate logging. If its logic was moved out to a `lib/**/*` file I'm not sure if it would just be code duplication. After further review, there are too many dependencies of these utils, so I can't safely refactor them at the moment. I would recommend considering the exclusive `acl` commands refactored and tackle the `dao` utils refactors latter.

## 🚨 Test instructions

Run `aragon dao acl` commands against your DAO

## ✔️ PR Todo

### DIFF
--- a/packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/knownRoles.js
+++ b/packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/knownRoles.js
@@ -10,7 +10,7 @@ const defaultAppsRoles = require('../../../../knownRoles.json')
  * @return {Object.<string, RoleDefinition>} Unique known roles
  */
 const getKnownRoles = module => {
-  const currentAppRoles = module.roles || []
+  const currentAppRoles = module ? module.roles : []
   const allRoles = defaultAppsRoles.concat(currentAppRoles)
 
   return keyBy(allRoles, role => keccak256(role.id))
--- a/packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/knownRoles.js
+++ b/packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/knownRoles.js
@@ -1,26 +1,19 @@
 const { keccak256 } = require('web3').utils
-const path = require('path')
-const { findProjectRoot } = require('../../../../util')
+const { keyBy } = require('lodash')
 const defaultAppsRoles = require('../../../../knownRoles.json')
 
-const currentAppRoles = () => {
-  try {
-    const arappPath = path.resolve(findProjectRoot(), 'arapp.json')
-    return require(arappPath).roles || []
-  } catch (_) {
-    return []
-  }
+/**
+ * Returns this app roles and default known roles
+ *
+ * TODO: add support for user apps
+ * @param {ArappConfig} module arapp.json contents
+ * @return {Object.<string, RoleDefinition>} Unique known roles
+ */
+const getKnownRoles = module => {
+  const currentAppRoles = module.roles || []
+  const allRoles = defaultAppsRoles.concat(currentAppRoles)
+
+  return keyBy(allRoles, role => keccak256(role.id))
 }
 
-// TODO: add support for user apps
-const rolesForApps = () => {
-  const allRoles = defaultAppsRoles.concat(currentAppRoles())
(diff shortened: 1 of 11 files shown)


mcormier:
Hey. I was mentioned above but this project does not remind me of anything.  Did I make a suggestion a long time ago that is now being considered or was the wrong person added in the description?  Cheers.

0xGabi:
Hi @mcormier sorry about that we have a team mate with a similar GitHub username

dapplion:
Yep, I mistaken you for @macor161. Thank you @0xGabi for amending the PR mentions.

dapplion:
@0xGabi In reference to all `acl` commands except `acl view`:
I would recommend considering the `acl` commands refactored. The next step would be to address the helpers and utils they depend on. However, after further review, there are too many dependencies of these utils, so it's not safe refactor them at the moment. It would be best to address all `dao` helpers/utils at once.

packages/aragon-cli/src/commands/dao_cmds/acl_cmds/utils/knownRoles.js
```diff
@@ -1,26 +1,19 @@
...
-  )
+/**
+ * Returns this app roles and default known roles
+ *
+ * TODO: add support for user apps
+ * @param {ArappConfig} module arapp.json contents
+ * @return {Object.<string, RoleDefinition>} Unique known roles
+ */
+const getKnownRoles = module => {
+  const currentAppRoles = module.roles || []
```
0xGabi:
```suggestion
  const currentAppRoles = module ? module.roles : []
```

packages/aragon-cli/src/lib/acl/viewFormatter.js
```diff
@@ -0,0 +1,86 @@
...
+  } = aclPermission
+
+  /**
+   * Shortcut to TRY get an app name from it's proxy address
+   * If the app can't be found, returns an empty string: ''
+   * @param  {string} address Proxy address of a possible app
+   * @return {AclPermissionAppInfo} app info
+   */
+  const getAppInfo = address => {
+    const app = apps.find(({ proxyAddress }) => address === proxyAddress)
```
0xGabi: I'm not sure why the kernel is printed with the address expression instead of `kernel (0x0..00)`
For what I could debug it is related with the address pass to this function be the `ZERO_ADDRESS` but couldn't understand why

dapplion: @0xGabi  could you provide steps to recreate, please?

0xGabi: Yes that's fine. I was testing with a new dao creating using `dao new` and for that particular case is not displayed like this.

packages/aragon-cli/src/lib/acl/typedef.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * @typedef {Object} PermissionsData
+ * @property {string[]} allowedEntities
+ * @property {string} manager
+ *
+ * ACL structure sample: {
```
0xGabi: @dapplion is this ACL structure sample complete?

dapplion: Until the `PermissionsData` level, I think so.

PR #865, opened 2019-11-02
Refactor deploy -  #858 task

dapplion:
# 🦅 Pull Request

- [x] deploy

Extracted the deploy logic into `lib/deploy`. Separated the original function into two: `linkLibraries` and `deployContract` where the first one is tested. Since `deployContract` is fairly simple and just interacts with web3 I don't see much value in testing it, where a stub style test will not provide much value. Please let me know if you think otherwise.

It would be nice to do a commit / PR merge to develop with typings for common arguments similar to what I started at https://github.com/aragon/aragon-cli/pull/864, so each refactor branch can benefit from it and DRY.

## ✔️ PR Todo


dapplion:
Thank you for reviewing it. I'll tackle the utils next

PR #870, opened 2019-11-05
Extracts logic from apm commands: extract-functions command

ajsantander:
_Ready for review_

Sub PR of #871

### DIFF
--- a/packages/aragon-cli/package.json
+++ b/packages/aragon-cli/package.json
@@ -89,11 +89,11 @@
   "devDependencies": {
     "@aragon/apps-agent": "^2.0.0",
     "@aragon/apps-finance": "^3.0.0",
+    "@aragon/apps-payroll": "^1.0.0",
+    "@aragon/apps-survey": "^1.0.0",
     "@aragon/apps-token-manager": "^2.0.0",
     "@aragon/apps-vault": "^4.0.0",
     "@aragon/apps-voting": "^2.0.0",
-    "@aragon/apps-payroll": "^1.0.0",
-    "@aragon/apps-survey": "^1.0.0",
     "@babel/cli": "^7.1.2",
     "@babel/core": "^7.1.2",
     "@babel/node": "^7.0.0",
@@ -119,7 +119,8 @@
     "prettier": "^1.15.3",
     "proxyquire": "^2.1.0",
     "regenerator-runtime": "^0.13.3",
-    "sinon": "^7.3.1"
+    "sinon": "^7.3.1",
+    "tmp": "^0.1.0"
   },
   "engines": {
     "node": ">=8.0.0"
--- a/packages/aragon-cli/src/commands/apm_cmds/extract-functions.js
+++ b/packages/aragon-cli/src/commands/apm_cmds/extract-functions.js
@@ -25,13 +25,7 @@ exports.handler = async function({ cwd, reporter, contract, output }) {
   const filename = path.basename(contractPath).replace('.sol', '.json')
   const outputPath = path.resolve(output, filename)
 
-  const progressHandler = step => {
-    switch (step) {
-      case 1:
-        reporter.success(`Saved to ${chalk.blue(outputPath)}`)
-        break
-    }
-  }
-
-  await extractContractInfoToFile(contractPath, outputPath, progressHandler)
+  await extractContractInfoToFile(contractPath, outputPath)
+  reporter.success(`Saved to ${chalk.blue(outputPath)}`)
+  process.exit()
 }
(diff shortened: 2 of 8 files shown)


packages/aragon-cli/src/helpers/solidity-extractor.js
```diff
@@ -84,21 +85,71 @@ const getRoles = declaration => {
...
+    return []
+  }
 
-  return funcDecs
-    .filter(dec => modifiesStateAndIsPublic(dec))
-    .map(dec => ({
-      sig: getSignature(dec),
-      roles: getRoles(dec),
-      notice: getNotice(dec),
+  const stateModifyingFunctionDeclarations = functionDeclarations
```
macor161: Much better names, I like that! 💯

packages/aragon-cli/src/helpers/solidity-extractor.js
```diff
@@ -84,21 +85,71 @@ const getRoles = declaration => {
...
+  // Extract all role ids from the function descriptors.
+  let roleSet = new Set()
+  functionDescriptors.forEach(({ roles }) => roles.forEach(role => roleSet.add(role)))
+  const roleIds = [...roleSet]
+
+  // Parse role ids into objects.
+  // TODO: Name and parameters are currently not being extracted, and it's probably better to get it from an AST instead of the Solidity code. For now, the properties are merely place holders.
+  return roleIds.map(id => ({
+    id,
+    bytes: '0x' + keccak256(id),
```
macor161:
```suggestion
    bytes: keccak256(id),
```

packages/aragon-cli/src/commands/apm_cmds/extract-functions.js
```diff
@@ -22,37 +20,18 @@ exports.builder = function(yargs) {
...
-  }
-
-  const filename = path.basename(contract).replace('.sol', '.json')
+exports.handler = async function({ cwd, reporter, contract, output }) {
+  const contractPath = path.resolve(cwd, contract)
+  const filename = path.basename(contractPath).replace('.sol', '.json')
   const outputPath = path.resolve(output, filename)
 
-  await writeJson(outputPath, content, { spaces: '\t' })
+  const progressHandler = (step) => {
```
macor161: Is the handler necessary? Is it in case we add more steps in the future?

packages/aragon-cli/package.json
```diff
@@ -89,11 +89,11 @@
   "devDependencies": {
     "@aragon/apps-agent": "^2.0.0",
     "@aragon/apps-finance": "^3.0.0",
+    "@aragon/apps-payroll": "^1.0.0",
```
dapplion: Avoid unnecessary changes in other sections of the package.json to prevent git conflicts

packages/aragon-cli/src/lib/apm/extractContractInfoToFile.js
```diff
@@ -0,0 +1,10 @@
+const extract = require('../../helpers/solidity-extractor')
+const { writeJson } = require('fs-extra')
+
+module.exports = async (contractPath, outputPath, progressHandler) => {
```
dapplion: What's the purpose of having a `progressHandler` that gets called once at the end? I would be cleaner to remove it and just log whatever the front-end needs after this async function has resolved.

packages/aragon-cli/src/commands/apm_cmds/extract-functions.js
```diff
@@ -22,37 +20,18 @@ exports.builder = function(yargs) {
...
-  }
-
-  const filename = path.basename(contract).replace('.sol', '.json')
+exports.handler = async function({ cwd, reporter, contract, output }) {
+  const contractPath = path.resolve(cwd, contract)
+  const filename = path.basename(contractPath).replace('.sol', '.json')
   const outputPath = path.resolve(output, filename)
 
-  await writeJson(outputPath, content, { spaces: '\t' })
+  const progressHandler = step => {
```
dapplion: See comment on https://github.com/aragon/aragon-cli/pull/870/files#diff-b43061eb5cc2ed67dbd7a209e622dca3R4

packages/aragon-cli/src/helpers/solidity-extractor.js
dapplion: Maybe we can do this in a safe way such as
```js
const hash = hashFunction(id)
return "0x" + (hash.split("0x")[1] || hash)
```
or
```js
const hash = hashFunction(id)
return hash.startsWith("0x") ? hash : "0x" + hash
```
If it's unclear what return format it does

macor161: @dapplion Yes I think it could be a good idea to abstract this away in case we switch again to another `keccak256` function that doesn't include `0x`. I will create a separate issue for this since the function is used at a lot of different places other than this PR.

macor161: See #914

packages/aragon-cli/package.json
macor161: Agree that we should avoid changes unrelated to the PR. I left it there because it will probably conflict with `develop` if I remove it.

PR #875, opened 2019-11-06
Refactor dao id assign logic

macor161:
# 🦅 Pull Request

Ref PR: #874

## 🚨 Test instructions

`dao id assign`

### DIFF
--- a/packages/aragon-cli/src/lib/dao/assign-id.js
+++ b/packages/aragon-cli/src/lib/dao/assign-id.js
@@ -1,0 +1,30 @@
+const { convertDAOIdToUrl } = require('../../util')
+const ENS = require('ethereum-ens')
+
+async function assignId(daoAddress, id) {
+
+
+}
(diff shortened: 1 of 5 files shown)


packages/aragon-cli/src/lib/dao/assign-id.js
```diff
@@ -0,0 +1,61 @@
+const { convertDAOIdToSubdomain, ARAGON_DOMAIN } = require('../../util')
+const ENS = require('ethereum-ens')
+const ififsResolvingRegistrarAbi = require('@aragon/id/build/contracts/IFIFSResolvingRegistrar')
+  .abi
+const { sha3, isAddress } = require('web3').utils
+const REGISTRAR_GAS_LIMIT = '1000000'
```
0xGabi: I wonder if we are able to use something like `getRecommendedGasLimit` 🤔

Issue #650, opened 2019-07-19
Replace coveralls with codecov


0x6431346e:
## 🚀 Feature

Coveralls can be very annoying sometimes reporting a coverage change when there are no code changes...perhaps we are not using it right, but https://codecov.io seems much nicer. E.g.: 
![image](https://user-images.githubusercontent.com/26041347/61535082-01691600-aa3a-11e9-989c-e454bcd177ef.png)

sohkai:
I've added aragonCLI to codecov and installed its Github app as well.

@0x6431346e are you able to access https://codecov.io/gh/aragon/aragon-cli?

0x6431346e:
> I've added aragonCLI to codecov and installed its Github app as well.
> 
> @0x6431346e are you able to access [codecov.io/gh/aragon/aragon-cli](https://codecov.io/gh/aragon/aragon-cli)?

Yes, I can access it, thanks!

Closed by #878 

I think the detailed report is great: https://github.com/aragon/aragon-cli/pull/877#issuecomment-551132650

PR #878, opened 2019-11-06
Improve build tools usage

0x6431346e:
# 🦅 Pull Request

Extracted some good changes from #709 :

- Moved `husky` to the root package so that it works for other packages like `create-aragon-app` not just `aragon-cli`
- Removed `lint-staged` (not sure if it's worth having it because linting all files instead of just the committed is still quite fast, what do you guys think? 0 it does seem to support a monorepo setup though: https://github.com/okonet/lint-staged#how-to-use-lint-staged-in-a-multi-package-monorepo)
- Paralleled some tasks that do not depend on each other like `lint`, `test`, `build`
- Moved `babel`, `eslint` and `prettier` configs to package.json (I personally find it much nicer because it does not pollute the file tree, thoughts?)
- Removed `documentation` as is very annoying and forces you to write jsdoc for each parameter once you start
- Removed `regenerator-runtime` as it's included by https://babeljs.io/docs/en/babel-preset-env
- Linted some tests files that were ignored

## ✔️ PR Todo


packages/aragon-cli/package.json
```diff
@@ -113,45 +111,62 @@
...
       "test/**/*.test.js"
     ]
   },
-  "aragon": {
-    "clientRepo": "https://github.com/aragon/aragon",
-    "clientVersion": "775edd606333a111eb2693df53900039722a95dc",
-    "clientPort": "3000",
-    "defaultGasPrice": "2",
-    "defaultIpfsTimeout": "300000"
+  "babel": {
```
0xGabi: Are we good with the configs we have for babel, eslint and prettier? Do we need to update something?

0xGabi:
>Moved babel, eslint and prettier configs to package.json

If there is no extra reason for keep them on a separate file, I agree, let's move them

0xGabi:
>Removed lint-staged

No strong preferences. I wonder, once the codebase move to typescript or grow more, having `lint-staged` could be a benefit?

0xGabi:
One extra detail, maybe makes sense to move to codecov as I saw Brett already configured: https://github.com/aragon/aragon-cli/issues/650

sohkai:
> Moved babel, eslint and prettier configs to package.json (I personally find it much nicer because it does not pollute the file tree, thoughts?)

No big preference, but it's likely this will be much cleaner once we have a shared project config (I keep promising this, but delivering soon).

> Removed documentation as is very annoying and forces you to write jsdoc for each parameter once you start

👍 We should just move to typescript, since we don't use the autogenerated doc anyway.

0x6431346e:
> LGTM
> 
> Do you think that make sense to include the types files that @dapplion defined on this PR?

I think we should aim for smaller PRs now that it isn't just the two of us working, it's easier to keep up with what/why changed and to reverse changes.

> No strong preferences. I wonder, once the codebase move to typescript or grow more, having lint-staged could be a benefit?

If the linting step start taking a long time ( I think now is 3-4 seconds) then yes because it would be annoying to wait a lot when you just want to make a temporary commit to not lose some changes.

> One extra detail, maybe makes sense to move to codecov as I saw Brett already configured: #650

:+1:

packages/aragon-cli/package.json
0x6431346e: I think they're fine so far, the only thing I added was:

```
          "targets": {
            "node": true
          }
```

which made the compile files a bit smaller and easier to read since we don't need to polyfill a bunch of stuff that browsers are missing.

0xGabi:
Thanks Daniel. And my bad for breaking the test with bumping a couple of packages. Will update the lock files and the merge it 💪

.github/main.workflow
```diff
@@ -1,4 +1,4 @@
-workflow "lint, test, build" {
```
sohkai: I was meaning to ask, we should update these to the yaml declaration instead; I don't these actions work anymore: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions

PR #871, opened 2019-11-05
Extracts logic from apm commands

ajsantander:
WIP

### Commands:
- [x] **[Merged]** apm extract-functions command - Sub PR: #870 
- [x] **[Merged]** apm info command - Sub PR: #879
- [x] **[Deprecated]** apm version command
- [x] **[Merged]** apm versions command - Sub PR: #891
- [x] **[Merged]** apm packages command - Sub PR: #892
- [x] **[Merged]** apm grant command - Sub PR: #902 
- [ ] _[WIP @dapplion]_ apm publish command -> **moved to next iteration**

### Misc:
- [ ] Tests have the same default values comment, they should all be /* Default values */
- [ ] Write tests for ACL.js
- [ ] Write tests for solidity-extractor.js
- [ ] Packages command does not need to import APM
- [ ] Most of the apm refactors are similar, especially in how they begin, can something be abstracted from them?
- [ ] Unify conversions from ens-registry to ensRegistryAddress in scripts and tests, or remove the need for the conversion altogether ) -> **will be done on removing deprecated code PR**
- [x] When removing tasks, lost the little green tick on console output titles... need a way to reproduce it on all the commands that I changed

Note: discussion about future steps moved #947

PR #879, opened 2019-11-06
Extracts logic from apm commands: info command

ajsantander:
_Ready for review_

Sub PR of #871


dapplion:
@0xGabi Sorry I had started the review but I forgot to publish it. I see this is merged now but I have raised relevant concerns about the PR.

PR #882, opened 2019-11-08
Update Github Action to new yaml syntax

0xGabi:
# 🦅 Pull Request


.github/workflows/main.yml
```diff
@@ -0,0 +1,19 @@
+on: push
+name: CI
+jobs:
+  CI:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install node
+      uses: actions/npm@master
+      with:
+        node-version: 12
```
sohkai: We may as well also add 8, 10, and 12 here if that's what we're also testing in Travis (see sample starter: https://github.com/actions/starter-workflows/blob/master/ci/node.js.yml#L12)

0x6431346e:
Ooops, sorry but I didn't see this PR and looked into this too: https://github.com/aragon/aragon-cli/pull/905

Do you wanna review that PR instead?

0xGabi:
Superseded by #905

PR #893, opened 2019-11-09
Refactor middlewares - #858 task

dapplion:
# 🦅 Pull Request

Extracted the middleware logic to two libraries 
- `lib/loadConfigFiles.js`: Finds and load JSON config files. For loading `arapp.json` and `manifest.json`
- `configEnvironment.js`: Pure function which receives options and config objects and returns the variables that were originally returned by the `environment` middleware: `apm`, `network`, `wsProvider`, and `arapp`.

Then, both libs are used in a single middleware file `middlewares/index.js` where it only parses options, delegates and adds cli-specific messages to the errors thrown by the libs. I have merged the previous three middlewares since `module` and `manifest` are now essentially one line of code, and this way we prevent a nasty mutation done before to circumvent how `yargs` treats middlewares. https://github.com/aragon/aragon-cli/blob/20ab9c05f3cbf35efd755bfb6670264c414d18ab/packages/aragon-cli/src/middleware/module.js#L26-L30

----

@0xGabi - **Note**: I have re-structured the entire logic of the **environment** middleware in order for it to be able to separate disk I/O, cli-specific errors and pure logic. I have done my best effort to ensure that the resulting logic **is equivalent** to the original, but please review it with extra attention to make sure it's strictly the same if necessay. I have only introduced two changes:
- 1. Removed `apmEnsRegistry` since it indeed appears to not be used by anyone
https://github.com/aragon/aragon-cli/blob/20ab9c05f3cbf35efd755bfb6670264c414d18ab/packages/aragon-cli/src/middleware/environment.js#L193-L194
- 2. Changed the order in which the `apm` object is merged, giving priority to the `apm` object passed as a parameter
https://github.com/aragon/aragon-cli/blob/20ab9c05f3cbf35efd755bfb6670264c414d18ab/packages/aragon-cli/src/middleware/environment.js#L181-L183

I've added a unit test suite to check what configurations `configEnvironment.js` returns given different flags and options. If possible, please suggest additional configurations any of you know that users are using.

---

**Note**: In order to make sense of the environment logic I wrote the lib files in typescript first. I've left the file in PR for the record as well as another typedef.js. Should I remove these files and add them later when we actually move to typescript?

## 🚨 Test instructions

Run any command (i.e. `dao`) that interacts with the blockchain and then use different `environment` and `network` options.

## ✔️ PR Todo


packages/aragon-cli/src/lib/configEnvironment.ts
```diff
@@ -0,0 +1,203 @@
...
+  const { provider, host, port } = truffleNetwork
+  return {
+    ...truffleNetwork,
+    name: network,
+    provider: provider
+      ? typeof provider === 'function'
+        ? provider()
+        : provider
+      : new Web3.providers.WebsocketProvider(
+          host && port ? `ws://${host}:${port}` : `http://localhost:8545`
```
0x6431346e: I'm a bit worried that this will not work with http providers

packages/aragon-cli/src/lib/configEnvironment.ts
```diff
@@ -0,0 +1,203 @@
...
+    return {
+      name: `frame-${network}`,
+      provider: new Web3.providers.WebsocketProvider(
+        FRAME_ENDPOINT,
+        { headers: { origin: FRAME_ORIGIN } } // Provider options
+      ),
+    }
+  }
+
+  const truffleNetwork = truffleConfig.networks[network]
```
0x6431346e: I get this:
```
/home/daniel/r/ar/aragon-cli/packages/aragon-cli/src/lib/configEnvironment.js:50
  const truffleNetwork = truffleConfig.networks[network]
                                       ^
TypeError: Cannot read property 'networks' of null
    at networks (/home/daniel/r/ar/aragon-cli/packages/aragon-cli/src/lib/configEnvironment.js:50:40)
```
when running `node dist/cli.js devchain`

packages/aragon-cli/src/lib/configEnvironment.ts
dapplion: When the user exclusively defines the `host` and `port` is there any other key to be used to guess the protocol?

packages/aragon-cli/src/lib/configEnvironment.ts
dapplion: @0x6431346e Thanks for testing it! I've committed https://github.com/aragon/aragon-cli/pull/893/commits/39303871c2b4256bb5e94e12f6bd1e4a9ecf9c52 which fixes the problem. I was able to reproduce the error locally and with this commit it works. Please if you can, confirm the issue is fixed in your local env.

> The issue was that the argument `truffleConfig` was `null` instead of `undefined` so it did not trigger the default value.

packages/aragon-cli/src/lib/configEnvironment.ts
0xGabi: Let's make sure to asses this in future tests

dapplion:
@0xGabi Thank you so much

PR #902, opened 2019-11-11
Extracts logic from apm commands: grant command

ajsantander:
_Ready for review_

Sub PR of #871


packages/aragon-cli/src/lib/apm/grantNewVersionsPermission.js
```diff
@@ -0,0 +1,76 @@
...
+    // the recommended gasLimit is already calculated by the ACL module
+
+    try {
+      const receipt = await web3.eth.sendTransaction(transaction)
+      if (progressHandler) {
+        progressHandler(4, receipt.transactionHash)
+      }
+    } catch (error) {
+      if (progressHandler) {
+        progressHandler(5, error)
```
dapplion: This is dangerous as the error could easily be ignored / not handled properly. The progressHandler should only deal with informative data updates, not stopping errors. Instead, I would advise using custom error classes https://javascript.info/custom-errors.

As example please look at this implementation
- Definition: https://github.com/aragon/aragon-cli/pull/893/files#diff-86adbef0eb8780f10a545fcba7442b0dR13
- Usage: https://github.com/aragon/aragon-cli/pull/893/files#diff-f8a8bf84d63a26b560eda2f22a89de9dR80

packages/aragon-cli/test/lib/apm/grantNewVersionsPermission.test.js
```diff
@@ -0,0 +1,293 @@
...
+  }
+})
+
+test.afterEach('cleanup', t => {
+  sinon.restore()
+})
+
+/* Tests */
+
+test('properly calls the progressHandler when something errors', async t => {
```
dapplion: See comment https://github.com/aragon/aragon-cli/pull/902/files#diff-786f4b2b7fe547319f5348c3316a396fR68

packages/aragon-cli/src/lib/apm/grantNewVersionsPermission.js
```diff
@@ -0,0 +1,76 @@
...
+  apmOptions,
+  grantees,
+  progressHandler,
+  { gasPrice }
+) => {
+  if (grantees.length === 0) {
+    throw new Error('No grantee addresses provided')
+  }
+
+  if (progressHandler) {
```
dapplion: Maybe we should only add `progressHandler` to where they provide valuable information. As a good rule of thumb, if they are not used by our front-end, they should not be there. And `progressHandler` before the first expensive async call or after the last do not provide much value either

packages/aragon-cli/src/lib/apm/grantNewVersionsPermission.js
macor161: I agree. I removed the error handling from the `progressHandler`. It will indeed probably be a good idea to use a custom error once we start focusing on the SDK behavior.

packages/aragon-cli/test/lib/apm/grantNewVersionsPermission.test.js
```diff
@@ -0,0 +1,285 @@
...
+  })
+
+  const grantNewVersionsPermission = proxyquire
+    .noCallThru()
+    .load('../../../src/lib/apm/grantNewVersionsPermission', {
+      '@aragon/apm': apmStub,
+      './util/acl': aclStub,
+    })
+
+  t.context = {
```
dapplion: We are trying to move away from the context pattern introduced by `yargs` that makes code more difficult to type are reason about. It would be best to declare the variables with `let` in the module scope, assign them here and then use them on each test.

packages/aragon-cli/src/lib/apm/grantNewVersionsPermission.js
dapplion: Thank you so much

dapplion:
Thank you @macor161 for addressing the issues, really neat work. I added one extra comment on a separate review regarding the tests

packages/aragon-cli/test/lib/apm/grantNewVersionsPermission.test.js
0xGabi: Agree, we can do that with all tests on the next iteration.

dapplion: Sounds good

PR #905, opened 2019-11-12
Fix GitHub Actions setup

0x6431346e:
# 🦅 Pull Request

## ✔️ PR Todo

- [ ] Include links to related issues/PRs


0x6431346e:
Looks like node v13 is failing as well as building on windows because of `node-gyp`

0xGabi:
I think I found a way to not depend on `node-gyp` anymore. We need to update to use ethers and use a rebuild of `ganache-core` or `ganache-cli` directly (cause they are already doing that.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,40 @@
+name: CI
+
+on: [push]
```
0xGabi: I'm afraid that the GitHub Actions could take to much time if we done them on every push. What you think of doing it only on develop and master?

PR #906, opened 2019-11-12
Extracts logic from apm commands: publish command

dapplion:
#### Plan for iteration v2
The code used in this command is massive and with abstractions that difficult moving the logic to `/lib`. Some amount of refactoring on the logic must be done in order to (1) properly separate the logic from the front-end, (2) move it to `/lib` and, (3) test it. Due to this, an e2e that completely covers the entire command is **mandatory** before moving forward. Also, ideally, the abstractions should be rethought to ease testing and moving the code to `/lib`. It will all depend on how much code @0xGabi and @ajsantander agree on changing.

- [ ] Add e2e tests for the main use cases of `apm publish` and make sure it works in `master` or `develop`.
- [ ] Move this interactive from util to a front-end file/function
https://github.com/aragon/aragon-cli/blob/5a3dc58447837fe06075749165e820651e4510e0/packages/aragon-cli/src/commands/apm_cmds/util/generate-artifact.js#L80-L89
- [ ] Rethink the utils/helpers to decouple logic from side-effects (file system read/writes) to ease unit testing.
- [ ] Move these atomic, logical, testable utils/helpers to `/lib`.
- [ ] (if applicable) Divide the publish tasks in smaller testable blocks, but bigger than single utils/helpers, Then, move then to `/lib`. 
- [ ] Move these atomic, logical, testable tasks to `/lib`.
- [ ] Successful fully tested refactor!

-----

#### Iteration 1 work

Subtask of #871.

- [x] Split publish.js for each task it contains
- [x] Move interactive input questions from utils / handlers
- [x] Move logic (utils / helpers) to `/lib`
- [ ] Unit test them


0xGabi:
superseded by #1214

PR #912, opened 2019-11-13
Refactor IPFS commands

0x6431346e:
# 🦅 Pull Request

Part of #858 

Extracted from #709 

The `detached=false` option on `aragon ipfs start` does not work.

## ✔️ PR Todo


0xGabi:
After an async conversation with Daniel, we agree that I'll review and merge this PR as it. He will continue working on it on next refactor

packages/aragon-cli/src/commands/ipfs.js
```diff
@@ -1,9 +1,17 @@
-const startCommand = require('./ipfs_cmds/start')
+import * as commands from './ipfs_cmds'
 
-exports.builder = function(yargs) {
-  return yargs.commandDir('ipfs_cmds')
-}
+export const builder = yargs =>
```
dapplion: Should we apply this pattern to all other subcommands? I like this more explicit approach than using `yargs.commandDir`, but is there a particular reason for it?

0x6431346e: Yes, because it affects the order `--help` is generated and I think we can add them in the order an user will most likely learn/use them.

It's a very small detail though so I wouldn't worry about changing the other commands, unless you're within that file or something :man_shrugging:

packages/aragon-cli/src/commands/ipfs_cmds/start.js
```diff
@@ -1,83 +1,133 @@
-const path = require('path')
-const TaskList = require('listr')
-const chalk = require('chalk')
-const {
-  startIPFSDaemon,
-  isIPFSCORS,
-  setIPFSCORS,
-  isIPFSRunning,
-} = require('../../helpers/ipfs-daemon')
+// import path from 'path'
```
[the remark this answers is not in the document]
0x6431346e: I did not finish re-writing the start command and I wanted to make sure I go through that logic again and not miss anything. 

What's left is to clean this and implement the `--detached=false` option, but at the same time I think we could merge it too :man_shrugging:

packages/aragon-cli/src/commands/ipfs.js
```diff
@@ -1,9 +1,17 @@
...
+    .command(commands.status)
+    .command(commands.view)
+    .command(commands.propagate)
+    .command(commands.uninstall)
+    .usage(`Usage: aragon ipfs <command> [options]`)
+    .demandCommand(1, 'You need to specify a command')
 
-exports.command = 'ipfs'
-exports.describe = 'Shortcut for `aragon ipfs start`.'
-exports.handler = startCommand.handler
```
0xGabi: Was this intended? If we remove this we are not able to use `aragon ipfs` as an alias of `aragon ipfs start`

0xGabi:
Will merge and we can iterate on how to handle the `aragon ipfs start` alias

packages/aragon-cli/src/commands/ipfs.js
dapplion: Sound good! Let's add this as a todo task for latter down the refactor road

packages/aragon-cli/src/commands/ipfs_cmds/start.js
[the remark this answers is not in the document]
dapplion: Sure

PR #950, opened 2019-11-23
WIP Syntax changes to boost code coverage

dapplion:
This PR has been opened to trigger `codecov` runs

- **Convert CommonJS to ES6 modules**. Absolutely no change was made to the logic of the files. The bulk of the changes were made leveraging the code transformation tool `cjs-to-es6` to assist in the process. Additionally, some tweaks were made:
  - In some commands, a variable `task` was renamed to `tasks` to resolve a scope conflict now that the exported function `task` is used as a normal variable instead of `export.task`
  - Import `web3` utils from `web3-utils` instead of from `web3` directly
  - Change `export default { libA, libB }` to `export const libA` and `export const libB`
  - In tests consuming modules with default exports via `proxyquire` the import was changed from `const fn = proxyquire('../fn')` to `const { default: fn } = proxyquire('../fn')`

### DIFF
--- a/packages/aragon-cli/src/cli.js
+++ b/packages/aragon-cli/src/cli.js
@@ -1,14 +1,13 @@
 #!/usr/bin/env node
 import 'source-map-support/register'
-const Web3 = require('web3')
+import Web3 from 'web3'
+import { configCliMiddleware } from './middleware'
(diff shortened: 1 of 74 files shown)


0xGabi:
Superseded by #995

PR #952, opened 2019-11-23
Add more integration tests

dapplion:
Add tests to increase the code coverage of the `/lib` directory.

As of https://github.com/aragon/aragon-cli/pull/952/commits/9e3bb5b18edc8352c542a4b30ca3e4da027dbef4 it increases the coverage of `/lib` by +32.68% to **73.64%**.

#### Tests never finish on error :heavy_check_mark: 
`travis` - `npm` - `ava` - `nyc` or someone in the chain seems to swallow the error when it fails, forcing the test to take extra 10 minutes for the Travis timeout to kick in.

I hadn't the time to investigate the cause yet.

#### No snapshot diffs :hourglass: 
AVA does not show a diff of snapshot from Travis CI. See example https://travis-ci.org/aragon/aragon-cli/jobs/616072516#L486

#### acl view :heavy_check_mark: 
This test always fails since ava thinks that the promise wrapping AragonJS never resolves, and fails with:
```
✖ Should get formated permissions for a dao apps
Promise returned by test never resolved
```
Trying different async constructions didn't solve the issue. Therefore the test is skipped until there's more time to find a solution

#### start :hourglass: 

According to https://github.com/aragon/aragon-cli/issues/858 the `start` commands are not yet refactored. However there's code in `/lib` but the code if highly coupled with the front-end calling `task.skip()` and mutating the `task.ctx`
@0xGabi, is this refactored? Is this a WIP?


0xGabi:
@dappLion The `start` command went the first command to be refactored a long time ago. I didn't tackle it yet, so if it is blocking you to keep testing feel free to take it

macor161:
I'll do the `start` command if you haven't already began working on it @dapplion

dapplion:
> I'll do the `start` command if you haven't already began working on it @dapplion

I can take it, which will ease testing it for me

0x6431346e:
> acl view
This test always fails since ava thinks that the promise wrapping AragonJS never resolves, and fails with:
> ✖ Should get formated permissions for a dao apps
> Promise returned by test never resolved
> Trying different async constructions didn't solve the issue. Therefore the test is skipped until there's more time to find a solution

Looks like the problem in our `getLocalWeb3` test utility: 
```diff
- new Web3.providers.HttpProvider(`http://localhost:8545`)
+ new Web3.providers.WebsocketProvider(`ws://localhost:8545`)
```

Because it was connecting using an http provider, the `wrapper.apps` observable emitted only once with `['Kernel']` and considering how `getApps` [works](https://github.com/aragon/aragon-cli/blob/develop/packages/aragon-cli/src/helpers/aragonjs-wrapper.js#L135) this promise would never finish.

0x6431346e:
> Tests never finish on error
travis - npm - ava - nyc or someone in the chain seems to swallow the error when it fails, forcing the test to take extra 10 minutes for the Travis timeout to kick in.

The issue seems to be that the devchain remains started and lerna is waiting for it to exit, which never happens.

It seems to be that this is the ["correct" npm behavior](https://github.com/npm/npm/issues/5493).

I found a solution in d76491a inspired by [this SO answer](https://stackoverflow.com/a/25292952/7793435) which forces the postscript to run even if the tests have failed and return the correct exit code to make the CI fail as it should.

dapplion:
Awesome, thanks for catching the issue

PR #944, opened 2019-11-21
WIP Refactor dao apps

macor161:
# 🦅 Pull Request

Ref PR: #874

## TODO

- [ ] Unit and integration tests

PR #955, opened 2019-11-25
Extract logic from "dao apps" and "dao new"

macor161:
# 🦅 Pull Request

Extract logic from `dao apps` and `dao new` to the `lib` folder.
Supersedes #944

Part of the integration tests are dependent on IPFS. They will therefore be added once IPFS is included.

## 🚨 Test instructions

`dao new`
`dao apps`


macor161:
hmm.. there is currently a weird Travis error on node 9 and 10. I'm looking into this.

packages/aragon-cli/src/commands/dao_cmds/new.js
```diff
@@ -113,76 +84,35 @@ exports.task = async ({
...
-            templateInstance ||
-            new web3.eth.Contract(abi, ctx.repo.contractAddress)
-
-          const newInstanceTx = template.methods[fn](...fnArgs)
-          const estimatedGas = await newInstanceTx.estimateGas()
-          const { events } = await newInstanceTx.send({
-            from: ctx.accounts[0],
-            gas: await getRecommendedGasLimit(web3, estimatedGas),
+        task: async ctx => {
+          console.log('registry: ', apmOptions.ensRegistryAddress)
```
0x6431346e: :wave:

packages/aragon-cli/src/lib/dao/new.js
```diff
@@ -0,0 +1,38 @@
...
+module.exports = async function({
+  repo,
+  web3,
+  templateInstance,
+  newInstanceMethod,
+  newInstanceArgs,
+  deployEvent,
+  gasPrice,
+}) {
+  const template =
```
0xGabi: I think this part is a bit awkward. Is been more and more clear that we need a clear distinction of how we handle the aragonPM information. Probably something to look a bit more in detail on current iteration after we merge this PR.

packages/aragon-cli/src/commands/dao_cmds/new.js
0xGabi:
```suggestion
```

macor161: Good catch thanks! 😄

packages/aragon-cli/src/lib/dao/new.js
macor161: I agree 💯
The `templateInstance` was originally used here https://github.com/aragon/aragon-cli/blob/develop/packages/aragon-cli/src/commands/run.js#L371 but we'll probably be able to simplify things once we start refactoring the `run` command.

packages/aragon-cli/src/commands/dao_cmds/new.js
macor161: Thanks! :)

dapplion:
According to AVA's documentation, most of the usage of `t.plan()` is not appropriate/necessary, see https://github.com/avajs/ava/blob/master/docs/recipes/when-to-use-plan.md

macor161:
> According to AVA's documentation, most of the usage of t.plan() is not appropriate/necessary

@dapplion Ah I didn't know that thanks! It is fixed now.

ajsantander:
@macor161 it seems that `dao new` is not working on this PR. It seems to hang on "Fetching template..."

macor161:
> @macor161 it seems that dao new is not working on this PR. It seems to hang on "Fetching template..."

@ajsantander hmm strange it's working here but something is clearly wrong then. I will review this furthermore.

macor161:
@0xGabi Thanks a lot!! 🙂

Issue #947, opened 2019-11-21
Refactor codebase for robustness


0xGabi:
## 🔮 Planning

1. Follow @0x6431346e's [proposal on how to add integration tests](https://github.com/aragon/aragon-cli/pull/904) and aim for **50-60%** of coverage by the end of the refactor. Focusing on the `/lib` folder, the logic.

2. Implement [common mocks](https://github.com/aragon/aragon-cli/pull/945) and patterns (e.g. `progressHandler`).

3. Move code logic away from the `wrapper` unless it’s absolutely necessary.

## Preliminary tasks

Diagram with current progress: [here](https://drive.google.com/file/d/1jqbzTfqwdLmd4lHOzzXpCoScRMpYNdHQ/view?usp=sharing)

- **Blue**: DONE
- **Yellow**: TODO
- **Red**: WIP

#### Review and merge
  - [x] https://github.com/aragon/aragon-cli/pull/912
  - [x] https://github.com/aragon/aragon-cli/pull/871
  - [x] https://github.com/aragon/aragon-cli/pull/946

#### Finish WIP on
  - [x] `reporter` middleware #965
  - [x] `token change-controller` #972
  - [x] https://github.com/aragon/aragon-cli/pull/874 
    - [x] `dao new` #955
    - [x] `dao act` #970 
    - [x] `dao apps` #955
  - [ ] `aragon start` #974
  - [ ] `aragon devchain` 
  - [ ] `aragon contracts` #971 
  - [ ] Add IPFS to integration tests #951

### Pre-work discussions:

- [ ] How to incorporate complex side effects into the `/lib` folder? These include: 
  - installing, configuring and running binaries: `ipfs`, `devchain` 
  - Usage of `npm`
  - File system manipulation
- [ ] How to test these complex side effects?

### Logic units that involved several commands

- [ ] `amp publish` **[ Dev: , Reviewers:  ]**  WIP PR: https://github.com/aragon/aragon-cli/pull/906 (Includes plan for iteration 2)
- [ ] `run` **[ Dev: , Reviewers:  ]** 
- [ ] `don't use process.exit` **[ Dev: , Reviewers:  ]**  proposal issue https://github.com/aragon/aragon-cli/issues/948
- [ ] `dao commands utils` **[ Dev: , Reviewers:  ]**
- [ ] `execHandler` **[ Dev: , Reviewers:  ]** 
[...]

PR #970, opened 2019-11-27
Extract logic from "dao act"

macor161:
# 🦅 Pull Request

Ref #874 and #947

## 🚨 Test instructions

`dao act`

## TODO

- [x] Refactor `execHandler`
- [ ] Test `execHandler`


ajsantander:
:warning: 

@macor161 I think that at this point, we should be able to at least isolate the call to execHandler in lib/ and write a test for it. Otherwise `dao act` ends up not being tested as a whole...

macor161:
@ajsantander Agreed, I will refactor it in this PR.

PR #971, opened 2019-11-28
Extract logic and fix "contracts" command

macor161:
# 🦅 Pull Request

- Fix broken `contracts` command
- Extract logic to `lib` folder
- Add tests

We should probably add more tests to this, but it's currently difficult because the command relies on `process.cwd()`. In a next iteration, we should add a `cwd` param so we can run it in the directory of our choice.

~~Another note: truffle returns a **code 1** for the help command, so it's basically considered as an error by `execa` and it throws an exception. We may want to detect this and handle it correctly in a future iteration.~~  -> Fixed.

## 🚨 Test instructions

`aragpn contracts`

### DIFF
--- a/packages/aragon-cli/src/commands/contracts.js
+++ b/packages/aragon-cli/src/commands/contracts.js
@@ -1,20 +1,12 @@
-const { runTruffle } = require('../helpers/truffle-runner')
+const { contracts, extractTruffleArgs } = require('../lib/contracts')
 
-exports.command = 'contracts'
+exports.command = 'contracts [command]'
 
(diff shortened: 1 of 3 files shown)


ajsantander:
@macor161 this command should be deprecated in v8, so I'm not sure if separating the logic is applicable anymore. However, I would definitely keep new tests around if they are included in the PR.

0xGabi:
Agree with @ajsantander we should prevent move code to `/lib`

macor161:
Agreed, closing this and I'm going to add the tests in a new PR

Issue #948, opened 2019-11-21
Don't use process.exit()


dapplion:
## 💥 Proposal

In relation to the current SDK v7 refactor push, the frequent use of `process.exit()` in random parts of the code makes it really difficult to do integration tests

As a good practice, a Node.js app should not rely on `process.exit()` to stop a script. Whatever is holding the event loop alive should be cleaned, if possible.

For example `dao apps` is not able to finalize on it's own. The npm package node debugger `wtfnode` states that this 3 items are holding the event loop alive:

```
[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 2 (tty) (stdio)
- Sockets:
  - 127.0.0.1:42572 -> 127.0.0.1:8545
```

**Proposal**:
1. Never use `process.exit(1)` for errors. Throw an error and then handle it at the top yargs level with appropriate styling and debug / verbose settings
2. Understand what's keeping the event loop alive, and address the issue directly, allowing to have command handlers that don't end with `process.exit(0)`

sohkai:
The file handles we should definitely be closing... but I'm not super sure how we'll be able to close the web3 connection. I guess we can go to `web3.currentProvider` and try to invoke `close()` on it (not sure if web3@1.x implements `close()` on the providers but I think so).

dapplion:
> The file handles we should definitely be closing... but I'm not super sure how we'll be able to close the web3 connection. I guess we can go to `web3.currentProvider` and try to invoke `close()` on it (not sure if [web3@1.x](mailto:web3@1.x) implements `close()` on the providers but I think so).

The first two file descriptors are integral parts of the node process so they are not a problem. Web3 1.x has indeed a way to close a websocket connection 
```
web3.currentProvider.connection.close()
```
The problem now is to unify an control the multiple providers that are floating around

dapplion:
After some research today I found this issues hanging the node process:

### Web3 providers
Web3 WebSocket providers can be gracefully closed with 
```js
web3.currentProvider.connection.close()
```
We should uniform having only one initialized Web3 object in a CLI command lifecycle and call `close()` after the command completes (in the hook described below).

Known initializations of Web3 providers:

https://github.com/aragon/aragon-cli/blob/d70b41b2e886522edbb32c71d443c92ca082ef0e/packages/aragon-cli/src/lib/configEnvironment.js#L43
[...]

PR #992, opened 2019-12-06
Remove or replace process.exit()

dapplion:
# 🦅 Pull Request

See https://github.com/dapplion/aragon-cli/pull/3 for an example of how tests can look like.

Partially addresses https://github.com/aragon/aragon-cli/issues/948.

**Motivation**: If a dev wants to test a command directly just remove the `process.exit()` at the end of the task function and let the hook finalize the app for you.

**Tasks**:
- [x] Use onFinishCommand hook. There are some open handlers that hang the event loop. While these issues and not addressed (yet), they will be forced to stop in one single place instead of in every command.
- [x] Remove the final command handlers process.exit() statements. With the `onFinishCommand` hook, this statements are redundant.
- [x] Replace the middle command handlers process.exit() statements with return. If the handlers return, the `onFinishCommand` hook will call process.exit()
- [x] Replace process.exit(1) statements with throw Error. The `.fail` Yargs hook will manage the error and call process.exit(1).

----

**Note**: The `yargs` dependency is pointing Github's master while they release it (should be in a few days)

### DIFF
--- a/packages/aragon-cli/package.json
+++ b/packages/aragon-cli/package.json
@@ -90,7 +90,7 @@
     "truffle-flattener": "^1.2.9",
     "web3": "^1.2.0",
     "which": "^2.0.1",
-    "yargs": "^15.0.1"
+    "yargs": "git://github.com/yargs/yargs.git#fe380cd356aa33aef0449facd59c22cab8930ac9"
   },
   "devDependencies": {
     "@aragon/apps-agent": "^2.0.0",
--- a/packages/aragon-cli/src/cli.js
+++ b/packages/aragon-cli/src/cli.js
@@ -35,7 +35,7 @@ const cli = yargs
   // blank scriptName so that help text doesn't display "aragon" before each command
   .scriptName('')
   .demandCommand(1, 'You need to specify a command')
-  .fail(AragonReporter.errorHandler)
+
   /**
    * OPTIONS
    */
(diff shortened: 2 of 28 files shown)


packages/aragon-cli/package.json
```diff
@@ -90,7 +90,7 @@
     "truffle-flattener": "^1.2.9",
     "web3": "^1.2.4",
     "which": "^1.3.1",
-    "yargs": "^13.2.4"
+    "yargs": "yargs/yargs"
```
0x6431346e: Is this correct?

dapplion: Yes, that's to install the current Github master since the version with the feature is not released yet. Maybe it would be better to lock it to a specific commit

packages/aragon-cli/src/cli.js
```diff
@@ -100,5 +100,15 @@ const cli = yargs
...
 AragonReporter.configure(cli)
 cli.middleware(MIDDLEWARES)
+
+// Runs if command.handler is successful
+cli.onFinishCommand(() => {
+  process.exit()
+})
+// Runs if command.handler throws
+cli.fail((...args) => {
+  AragonReporter.errorHandler(...args)
```
ajsantander: Shouldn't we use process.exit(1) here?
(And process.exit(0) in line 105?)

ajsantander:
LGTM.
Great work @dapplion !!

packages/aragon-cli/src/cli.js
dapplion: 1) It's not necessary to specify `0`, according to the docs, the default is 0 https://nodejs.org/api/process.html#process_process_exit_code
2) The `AragonReporter.errorHandler` handles the error reporting and calls process.exit(1) when necessary

PR #993, opened 2019-12-06
Adds a test for the Custom Deploy guide

ajsantander:
The guide in question is https://hack.aragon.org/docs/guides-custom-deploy

This test simply reproduces all the commands called in the guide, and the test passes if no command fails and everything works as expected.


packages/aragon-cli/test/guides/customDeploy.test.js
```diff
@@ -0,0 +1,286 @@
...
+  // Identify accounts
+  const web3 = await getLocalWeb3()
+  const accounts = await web3.eth.getAccounts()
+  ACCOUNT1 = accounts[0]
+  ACCOUNT2 = accounts[1]
+  if (verbose) console.log(`ACCOUNT1`, ACCOUNT1)
+  if (verbose) console.log(`ACCOUNT2`, ACCOUNT2)
+})
+
+test.serial('creates a dao', async t => {
```
0x6431346e: Consider using `import { serial as test } from 'ava'` and then `test('...'`
I find it cleaner and easier to reason about: you know all tests are serial.

packages/aragon-cli/test/utils.js
```diff
@@ -91,3 +95,16 @@ export async function assertIpfsIsInstalled() {
     throw Error(`IPFS is not installed: ${e.message}`)
   }
 }
+
+export function matchAddressAtLineContaining(str, query) {
+  function matchAddress(str) {
```
0x6431346e: Found this confusing first time I read it, I assumed these functions are declared here because they are being returned. Perhaps we should avoid nested function declarations?

PR #995, opened 2019-12-08
Update to use ES6 modules

0xGabi:
# 🦅 Pull Request

Build on #950 after refactor new package.
Add `aragen` commands to `aragon-cli`.


packages/aragon-cli/src/commands/dao_cmds/utils/getRepoTask.js
```diff
@@ -2,7 +2,7 @@ import { DEFAULT_IPFS_TIMEOUT } from '@aragon/toolkit/dist/helpers/constants'
 //
 const LATEST_VERSION = 'latest'
 
-module.exports = {
+export default {
```
dapplion: **Todo**: Replace 
```js
export default { funcA, funcB }
```
with
```js
export funcA
export funcB
```
**And** make sure other modules importing it are compatible with the new syntax

packages/aragon-cli/src/commands/devchain_cmds/utils/ganache-vars.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
dapplion: **Todo**: Replace 
```js
export default { funcA, funcB }
```
with
```js
export funcA
export funcB
```
**And** make sure other modules importing it are compatible with the new syntax

dapplion:
@0xGabi Thanks for the quick fix of the suggested changes!

I propose a small improvement in the naming of the functions now that we are editing these files, for future clarity https://github.com/aragon/aragon-cli/pull/1007

packages/aragon-cli/package.json
```diff
@@ -1,188 +1,189 @@
...
-    ],
-    "parser": "babel-eslint"
-  },
-  "prettier": {
-    "singleQuote": true,
-    "semi": false,
-    "trailingComma": "es5",
-    "bracketSpacing": true
-  }
+	"name": "@aragon/cli",
```
0x6431346e: Any reason for converting this spaces into tabs?

0xGabi: Mmm no,we should update that 😅

PR #1018, opened 2019-12-10
Unskip tests

ajsantander:
Un-skip all skipped tests ftw!


packages/aragon-cli/src/commands/start.js
```diff
@@ -16,7 +16,7 @@ const DEFAULT_CLIENT_REPO = pkg.aragon.clientRepo
 const DEFAULT_CLIENT_VERSION = pkg.aragon.clientVersion
 const DEFAULT_CLIENT_PORT = pkg.aragon.clientPort
 
-exports.command = 'start [client-repo] [client-version]'
+exports.command = 'start [client-repo] [client-version] [openInBrowser]'
```
dapplion: We should not add a minor option in the command description

packages/aragon-cli/test/start/start.test.js
```diff
@@ -1,25 +1,31 @@
...
+    }
+  })
+
+  try {
+    await subprocess
+  } catch (err) {
+    if (!testPassed) {
+      t.fail()
+    }
+  }
```
dapplion: This is really ugly, but if it works...

¯\\_(ツ)_/¯

PR #1026, opened 2019-12-11
Minimize mocking in toolkit tests

ajsantander:
Status: Ready for review 

## Rationale

Right now, a number of tests in the toolkit package are mocking a bit too much. This is ok for pure unit tests, but we want our tests to be more like integration tests.

For example, getApmRegistryPackages.test.js doesn't use a real version of apm.js, but mocks it. As a result, the test "anchors" the implementation of getApmRegistryPackages.js but doesn't really test if the function produces the expected results in a real environment.

This PR minimizes the use of mocking in the tests, shifting them from pure unit testing to integration testing, reducing the coverage of tests, but concentrating the testing on critical aspects of the toolkit package.

## Tests to be ported (toolkit)
- [x] apm
  - [x] util/acl.test.js
  - [x] getApmRegistryPackages.test.js
  - [x] getApmRepo.test.js
  - [x] getApmRepoVersions.test.js
  - [x] grantNewVersionsPermission.test.js
- [x] dao
  - [x] apps.test.js
  - [x] assign-id.test.js
  - [x] encodeCall.test.js
  - [x] new.test.js
- [x] helpers
  - [x] aragonjs-wrapper.test.js
- [x] ipfs
  - [x] config.test.js
  - [x] daemonAndConfig.test.js
  - [x] data-viz.test.js
  - [x] data.test.js
  - [x] propagation.test.js
- [x] token
  - [x] factory.test.js

## Additional tasks
- [x] Remove mock helpers
- [x] No skipped or onlyed tests in toolkit


macor161:
The PR is ready for review. Two things to note:
- Mocking has been removed everywhere except for `getAppKernel`, I think this function is a good use case for it. What do you guys think?
- Accessing the `description` property here https://github.com/aragon/aragon-cli/pull/1026/files#diff-d821718085893309b36ee3225dcb5b7dR72 throws a pretty weird exception, only on nodejs 10 so I commented the assertion for the moment.

PR #1038, opened 2019-12-13
Bundle toolkit compilation output

0x6431346e:
# 🦅 Pull Request

Currently we are using the toolkit like this:
```
const {
  convertStringToParam,
  encodeParam,
} = require('@aragon/toolkit/dist/acl/params')
```

I think this could be problematic because we can't move files around any longer, if we move a function around the API breaks and we need to push a new major version.

With this change, we are exposing a single module, a large concatenated file, and people then need to use this syntax:

```
const {
  convertStringToParam,
  encodeParam,
} = require('@aragon/toolkit')
```

This gives us freedom over the internal structure of the toolkit.

## ✔️ PR Todo

- [x] Export all functions in `toolkit/src/index.js`
- [x] Fix and finish to expose the whole `toolkit` package


0x6431346e:
Depends on https://github.com/aragon/aragon-cli/pull/995
Other than that ready to be merged.

macor161:
Nice! 💯

ajsantander:
@0x6431346e can we rebase on develop and make sure all tests are passing before merge?

0x6431346e:
Thanks @0xGabi for picking this up :heart:

PR #1039, opened 2019-12-13
Add typescript & typedoc

### LINKED ISSUE #1045: Migrate toolkit to typescript
TBD

0x6431346e:
# 🦅 Pull Request

- Add `typescript` to close #1045
- Add `typedoc`, see the docs: https://cdn.statically.io/gh/aragon/aragon-cli/05d56998baad7927eefc1f3e8da0d3f237b3460d/packages/toolkit-typed/docs/index.html

~Node: if we decide we want this, it might make sense to freeze `toolkit` and rename all its files to `.ts`, rather than having two packages which can complicate things.~

Update: renaming all files to `.ts` is no longer necessary. Just like #1116 " .js and .ts will coexist while we progressively transition from one to the other." thanks @dapplion :pray: 

This setup uses typescript *through* babel because:
- It allows us to use babel plugins (new js features and third-party plugins like: [improve lodash imports](https://github.com/lodash/babel-plugin-lodash), [enhance console.log](https://github.com/mattphillips/babel-plugin-console), or [strip console.log](https://github.com/betaorbust/babel-plugin-groundskeeper-willie)
- It easier to maintain one compiler, this setup is more consistent with our other packages
- It's faster to compile (and in watch mode), because you can run `tsc` in a different process:

```sh
npm run build  1,83s user 0,15s system 156% cpu 1,263 total
npm run check-types  5,21s user 0,14s system 213% cpu 2,509 total
```

vs

```sh
npm run build  3,84s user 0,11s system 190% cpu 2,079 total
```

This also means you can check for type errors only when you’re ready, so if you're trying a hacky solution the compiler won't get in your way.

:point_up_2: This is summary of the following article: https://iamturns.com/typescript-babel/

## ✔️ PR Todo


0x6431346e:
> * In this "demo" files of the typescript setup, `eslint-disable` is used 5 times already. If we are using eslint and typescript is good to conform as much as possible to the enforced standards and this overuse from the get-go may be sending the wrong message to other devs down the line

The use of eslint-disable is mostly justified. If you have code change to suggest please do so.

> Also I'm skeptical of allowing this (below) to happen. Most of us are not used to using Typescript consistently so this can eventually be counterproductive incentivizing to produce untyped or badly typed code. Please, if you think otherwise let me know, this is very subjective.

Just like we do with linting, we allow the developer the freedom to worry about types when he wants to, but always before committing or before merging. I think this is preferable. For whomever wants to be ensuring type safety from the get-go, they can run `npm run lint` after each build, or better yet, intregate eslint into their IDE as you said, and fix the types right away. I don't see why a specific workflow has to be enforced, when it can be opt-in.

>Typescript is really powerful when it's integrated into your editor, so you don't even need a file watcher + compiler to check type safety, it can happen as you type.

Exactly. Therefore there's no need to include type checking during `build:watch` and slowing it down.

0xGabi:
Hey @0x6431346e thank you so much for keep pushing for typescript. I wonder if you could rebase to `next` so we can keep working on that branch until we release a new version for cli v7 🙏

PR #1061, opened 2019-12-14
Boilerplates CI and Tests fixes

0xGabi:
# 🦅 Pull Request

Small cosmetic changes on tests. 

Fix `startProcess` on toolkits to make it work with CI on the boilerplates.

Move `run`, `publish` and `start` tests to boilerplate.

Solve issue with `askForConfirmation` promise.


packages/aragon-cli/test/start/start.test.js
```diff
@@ -1,32 +0,0 @@
-import test from 'ava'
-import execa from 'execa'
-
```
ajsantander: If this test is removed, then are we testing the start command anywhere else?

0xGabi: Yep, I move it to the boilerplates

packages/toolkit/src/node/process.js
```diff
@@ -60,7 +63,9 @@ export const startProcess = async ({
       if (data.includes(readyOutput)) {
         resolve({
           output,
-          kill: () => killProcessTree(subprocess, { logger }),
+          kill: () => {
```
0x6431346e: We need to return `killProcessTree` here so that we can `await` and get notified when this function has finished. Let's revert this

packages/toolkit/test/token/token.test.js
```diff
@@ -41,8 +40,6 @@ test('deployMiniMeTokenFactory: should deploy the contract with the right args',
 })
 
 test('changeControler: should set the correct controller', async t => {
-  t.plan(2)
```
0x6431346e: Any reason for deleting the `t.plan` statements ?

0x6431346e:
> LGTM! Thanks for removing the unnecessary `t.plan()`

I personally like them because I used to get a lot of false positives, tests passing but no assertions run, when testing `rxjs` code. 
We might not need them with `async/await` code, but there might be places where this is useful! Let's not rush into removing them.

packages/toolkit/src/node/process.js
0x6431346e: Fixed in https://github.com/aragon/aragon-cli/pull/1072

PR #1086, opened 2019-12-17
Adds tests for the agent guide, dao new, and fixes a bug on act

ajsantander:
An await was missing on act :)

Adds 10 tests for the agent guide.
Adds 1 test for dao new with templates.


packages/cli/src/commands/dao_cmds/act.js
```diff
@@ -66,7 +66,7 @@ export const handler = async function({
 
   reporter.debug('Encoded call data: ', encodedCallData)
 
-  const task = execHandler({
+  const task = await execHandler({
```
macor161: Good catch!! 🙌

Issue #1090, opened 2019-12-17
Commands failing with --env aragon:rinkeby


ajsantander:
@dapplion @0xGabi 

I identified what causes this problem, but don't have enough context to implement a proper fix.

If I change [line 96 in configEnvironment.js](https://github.com/aragon/aragon-cli/blob/develop/packages/cli/src/lib/environment/configEnvironment.js#L96) to:
```
    apm: {
      ...merge(apm, env.apm),
      'ens-registry': env.registry,
    },

```

 I can succesfully run a command like `aragon apm packages --env aragon:rinkeby`, but tests break all around. So, I'm guessing that we need to fix that, but properly :)

We should also identify the pattern of the error and try to look for similar errors performed during logic extraction.

0xGabi:
Perfect, thanks @ajsantander for find that out

0xGabi:
Solved with #1100

PR #1100, opened 2019-12-18
Fix envrionments adding configure APM function

0xGabi:
# 🦅 Pull Request

Solves #1090 

Fix interaction with all environments. 

Use a `configureAPM` function to setup the global apm option.


packages/cli/src/lib/environment/configEnvironment.js
```diff
@@ -93,12 +118,11 @@ export function configEnvironment({
     : null
 
   return {
-    apm: {
-      'ens-registry': env.registry,
-      ...merge(env.apm || {}, apm || {}),
-    },
-    // if there is no arapp.json and the command is not init default to the "global" config
-    // designed for the dao commands including dao acl
+    apm: configureAPM(
```
0xGabi: Here is where the magic happen 🔮

packages/cli/src/lib/environment/configEnvironment.js
```diff
@@ -30,6 +34,25 @@ function getEnv(arapp, environment) {
   }
 }
 
+const configureAPM = (ipfsRPC, gateway, ensRegistryAddress) => {
```
dapplion: Please add JSDocs to this function to known the data structures of `ipfsRPC, gateway, ensRegistryAddress`. If not full at least parcial :pray:

0xGabi: Let's do it after release

dapplion: Sounds good

PR #1105, opened 2019-12-19
Run tests without subprocess and split tests and examples

dapplion:
# 🦅 Pull Request

Builds on https://github.com/aragon/aragon-cli/pull/1099.

- Instead of calling a node script via a subprocess, the example guide is run directly within node, which provides better debugging and clarity.
- Test runs are divided into two NPM scripts `npm run test` and `npm run test:integration`. It can make sense to not run the examples test on the git hooks, and run them only on CI. We could also skip coverage for the guides / examples with a similar technique

## 🚨 Test instructions

```
npm run test:integration
```


packages/toolkit/test-examples/createSingleVote.test.js
```diff
@@ -0,0 +1,119 @@
...
+      web3,
+      provider: web3.eth.currentProvider,
+      ipfs: {
+        rpc: { protocol: 'http', host: 'localhost', port: 5001, default: true },
+        gateway: 'http://localhost:8080/ipfs',
+      },
+      registryAddress: ensRegistryAddress,
+      ensRegistryAddress,
+      'ens-registry': ensRegistryAddress,
+    }
```
0xGabi:
```suggestion
	const options = {
      web3,
      provider: web3.eth.currentProvider,
      apm: {
        ipfs: {
	      rpc: { protocol: 'http', host: 'localhost', port: 5001, default: true },
	      gateway: 'http://localhost:8080/ipfs',
	    },
      	ensRegistryAddress: '0x5f6f7e8cc7346a11ca2def8f827b7a0b612c56a1',
      }
    }
```

packages/toolkit/test-examples/createSingleVote.test.js
```diff
@@ -0,0 +1,119 @@
...
+
+    // Create a single vote with all the other votes encoded in the script.
+    console.log(`Creating bundled vote with EVM script...`)
+    const tx = await exec({
+      web3,
+      dao: daoAddress,
+      app: votingAddress,
+      method: 'newVote',
+      params: [script, 'Execute multiple votes'],
+      apm: options,
```
0xGabi:
```suggestion
      apm: options.apm,

File: packages/toolkit/test-examples/createSingleVote.test.js [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+      ensRegistryAddress,
+      'ens-registry': ensRegistryAddress,
+    }
+
+    // Retrieve DAO template from APM.
+    const repo = await getApmRepo(
+      web3,
+      'membership-template.aragonpm.eth',
+      'latest',
+      options
```
0xGabi:
```suggestion
     options.apm

dapplion:
> The `option` object was a bit awkward. Also in the example only the `apm` configuration was utilized. Do we want to remove the provider and web3?

@0xGabi I agree, the `option` is awful but that's what's required right now to make it work. This should be used a hint to improve the interface / API of the toolkit.

Issue #1135, opened 2019-12-26
Fix skipped and failing tests


macor161:
The `create-aragon-app: create-app > should create a new aragon app based on the react boilerplate` is currently failing and the following tests are skipped:

```
@aragon/cli:   - app › app › should publish an aragon app directory successfully
@aragon/cli:   - app › app › should fetch published versions to aragonPM
@aragon/cli:   - app › app › should run an aragon app successfully on IPFS
@aragon/cli:   - app › app › should run an aragon app successfully on IPFS using a Template
@aragon/cli:   - app › app › should run an aragon app successfully on HTTP
@aragon/cli:   - app › app › should run an aragon app successfully on HTTP using a Template
```
We should fix these tests.

PR #1164, opened 2020-01-03
Fix skiped tests

0xGabi:
# 🦅 Pull Request

Solves #1135 

The test are working when ran standalone. There is still issues with the app tests.


macor161:
Superseded by #1229

PR #1201, opened 2020-01-13
Buidler plugin v0.1

### LINKED ISSUE #1056: [EPIC] Buidler plugin v0.1
aragon-cli currently supports developer oriented commands like start, run, devchain, etc.

These commands should be moved out of the CLI, into a Buidler plugin. The CLI will no longer be developer oriented and instead focus on power users. The Buidler plugin will be 100% developer oriented.

### Sprint 1
- [x] Define specification for Buidler plugin v1.0 #1162
- [x] Prototype fast buidler plugin start command #1178
- [x] Buidler prototype with working front/back end relading in the browser #1200
- [x] Bug: Back-end watcher triggers on start. It doesn't need to.
- [x] Full code review with Typescript style, including a linter set up.
- [x] Consistent inline docs.
- [x] Minimal test coverage. #1209
- [x] Config file that holds global values. #1212
- [x] Convert into an actual plugin. #1215 - ETA: Monday 20th, after pair programming session with Pato.
- [x] No hard coding to counter app.
- [x] Introduce prettier (tslint already being used). Should we follow Buidler style or Aragon style?
- [x] Review front end watching. #1216
- [x] ~Options to specify port, etc.~ Might not want these, since we are currently able to modify options via buidler.config.ts. Options are hard to remember and often overwhelm.
- [x] Design and improve console output. Tricky. We want relevant info to stick around, such as DAO address, client browser link. We also want to see what's going on with plenty of info. Should we use colors? Two windows? Listr (I know we didn't want it before, but it might give us the solution we want here)?
- [x] Proper generation of app artifacts #1206 *Wait for migration of generate-artifacts from cli to toolkit*
- [x] Review console output. E.g. there is a rogue console.log in repo.ts. Do we want a rule to disallow console logging except in the logger?
- [x] Test start task #1238 
- [x] Review the minimal amount of files that need to be at the plugin level. Do we need artifacts/, typechain/ or even contracts/?
- [x] Review the minimal tsconfig settings that we need.
- [x] Review npm dependencies at the plugin level and in buidler-project. Are we using lodash? Do we need to import ts-node at the plugin level, or any level at all? Do we want to have the toolkit dependency if we're not using it yet?
- [x] Resolution of issue in truffle-typings: https://github.com/ethereum-ts/truffle-typings/issues/14. We're currently using the workaround I mention in that issue, which implies making manual changes to node_modules/truffle-typings/
- [x] Reconsider method to keep start task open using Promise.all()
- [x] Rebase on latest develop to fix aragen issues. _Waiting for changes on develop..._
- [x] Install app/ dependencies (npm i) on main install. Wait for #1215 

### To address in Sprint 2
- Fail loudly when another process is using one of the target ports. Right now it fails very silently, and very awkwardly.
- Fix import of toolkit in build.ts.
- Move to standalone repo.
- No skipped tests. #1217
- On push, lint and test and don't allow if there are any errors.
- Set up TravisCI.
- Set up npm package.
- Only throw BuidlerPluginErrors: https://buidler.dev/advanced/building-plugins.html#throwing-errors-from-your-plugins
- Separate npm dependencies and dev-dependencies. Also consider: https://buidler.dev/advanced/building-plugins.html#throwing-errors-from-your-plugins - Do #1215 first.
- Add license and README. Follow this style: https://github.com/nomiclabs/buidler/tree/master/packages/buidler-etherscan
- Study one of the official Buidler plugins and make sure that we are following their conventions/styles, etc. Depends on #1215
- Add watchers for app artifacts https://github.com/aragon/aragon-cli/issues/1230
- Hot reloading: review why build is needed, if we're just watching. Is there a shorter/faster path?
- Hot reloading works on script.js, but does not automatically trigger a refresh in the browser.
- The app doesn't show until both builds are complete, and sometimes, if it takes long, the browser doesn't refresh after the builds are done. A manual refresh does the trick. When this happens, it feels a bit flaky.
- Review watch mechanism so that we don't depend that much on the project in app/.
- Ability to override certain parts of the plugin's start command pipeline.
- Ability to specify a template with which to deploy an app.
- Ability to develop Templates as well as Apps.
- Package this into create-aragon-app. An alternative would be to add commands to the plugin to build a configurable project layout with inquirer prompts. Or both.
- Use Buidler EVM with snapshots, instead of aragen.
- Ability to publish packages to apm, using the toolkit.
- ...

ajsantander:
Fix #1056 

This PR presents a prototype for a Buidler plugin, with a single `npx buidler start` command, with no options (has a config file tho), that allows for Aragon app development with hot reloading on the front end as well as the back end (smart contracts + script.js).

DEMO (click to play video):
[Plugin demo](https://www.youtube.com/watch?v=9QG93zjVXGM&feature=youtu.be)


ajsantander:
Closing PR, since we won't be merging this into aragon/aragon-cli, but instead move it to its own repo at aragon/buidler-aragon.

PR #1222, opened 2020-01-20
Move logic #3

0xGabi:
# 🦅 Pull Request

Build on #1214 

Refactor generate artifact logic. Transform it into a pure function and move it to the toolkit. 
During refactor I notice there were a few keys that were not needed anymore (`env`, `appName`) and decided to use each environment to generate information like `appId`. 

Note: The `deployment` key was also removed following discussion about supporting a unique IPFS hash across environments. There were values nondeterministic.

Tasks to address on a new PR:

- Add test for pure function

- Add logic for `deprecatedFunctions`

### DIFF
--- a/packages/cli/src/commands/apm_cmds/util/runPrepareForPublishTask.js
+++ b/packages/cli/src/commands/apm_cmds/util/runPrepareForPublishTask.js
@@ -1,6 +1,14 @@
 import tmp from 'tmp-promise'
 import TaskList from 'listr'
-import { APM_INITIAL_VERSIONS, apmPublishVersionIntent } from '@aragon/toolkit'
+import path from 'path'
+import fs from 'fs'
+import { promisify } from 'util'
+import { readJsonSync } from 'fs-extra'
+import {
+  APM_INITIAL_VERSIONS,
+  apmPublishVersionIntent,
+  generateApplicationArtifact,
+} from '@aragon/toolkit'
 //
 import listrOpts from '../../../helpers/listr-options'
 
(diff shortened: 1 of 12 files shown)


packages/cli/src/commands/apm_cmds/util/runPrepareForPublishTask.js
```diff
@@ -1,21 +1,29 @@
...
   changeManifestForHttpServedFrom,
-  generateApplicationArtifact,
   writeApplicationArtifact,
   flattenCodeFileExists,
   generateFlattenedCode,
   artifactExists,
   checkIfNewArticatIsIdentical,
   copyCurrentApplicationArtifacts,
 } from '../../../lib/apm/generateArtifact'
+const readFile = promisify(fs.readFile)
```
macor161: We can now use `fs.promises` since we don't support nodejs 8 anymore.

```suggestion
const readFile = fs.promises.readFile
```

packages/cli/src/commands/apm_cmds/util/runPrepareForPublishTask.js
```diff
@@ -113,25 +120,34 @@ export default async function runPrepareForPublishTask({
           }
 
           async function invokeArtifactGeneration() {
-            const {
-              artifact,
-              missingArtifactVersions,
-            } = await generateApplicationArtifact(
-              web3,
+            // const {
```
macor161: We can probably remove this commented code.

packages/toolkit/src/helpers/extractContractInfoToFile.js
```diff
@@ -1,9 +1,15 @@
+import fs from 'fs'
+import { promisify } from 'util'
 import { writeJson } from 'fs-extra'
 //
 import { extractContractInfo } from './solidity-extractor'
+const readFile = promisify(fs.readFile)
```
macor161: We also probably want to use `fs.promises` here instead of `promisify`

packages/toolkit/src/helpers/generateArtifact.js
```diff
@@ -0,0 +1,93 @@
+import namehash from 'eth-ens-namehash'
```
macor161: Nice clean up! 😄

PR #1226, opened 2020-01-20
Buidler plugin

dapplion:
# 🦅 Pull Request

Progress the work for a working buidler plugin with the aragon command start
- Logging solution like the one used by docker-compose where the logs are prepended by a color coded tag of their origin (in our case `back` and `front`)
- Simplify front-end watching since we are using parcel's server with watch. I suggest we should add one more watcher to the `arapp.json` file and trigger the building of artifacts when:
  - `*.sol` source code changes
  - `arapp.json` changes
- Make setPermissions dynamic from arapp.json
- Setup prettier and run lint + prettier on all files

### DIFF
--- a/packages/buidler/.prettierignore
+++ b/packages/buidler/.prettierignore
@@ -1,0 +1,6 @@
+cache/
+dist/
+typechain/
+buidler.config.ts
+# Temp, CounterApp contents
+app/
--- a/packages/buidler/.prettierrc
+++ b/packages/buidler/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "semi": false
+}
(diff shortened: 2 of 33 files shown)


ajsantander:
Fantastic work @dapplion ! Thanks :raised_hands: 

Could you please turn this into an issue?
 

> I suggest we should add one more watcher to the arapp.json file and trigger the building of artifacts when:
> *.sol source code changes
> arapp.json changes

dapplion:
> Fantastic work @dapplion ! Thanks 
> 
> Could you please turn this into an issue?
> 
> > I suggest we should add one more watcher to the arapp.json file and trigger the building of artifacts when:
> > *.sol source code changes
> > arapp.json changes

Sure https://github.com/aragon/aragon-cli/issues/1230

PR #1265, opened 2020-01-23
Use an existing package as static webserver

dapplion:
# 🦅 Pull Request

Part of buidler plugin codebase review.

Follow @macor161 to not use a self-written webserver in favor of an existing library.


dapplion:
Closing for https://github.com/aragon/buidler-aragon/pull/2

PR #1294, opened 2020-01-29
Refactor toolkit interfaces

0xGabi:
# 🦅 Pull Request

Split #1236 to only include changes to the Toolkit. The test for the Toolkit are passing and the next step is change aragonCLI to use the new interfaces


ajsantander:
The changes to the interface look awesome :heart_eyes: 
Taking a detailed look now...

If this is merged, since the toolkit and the cli are in the same repo, the cli will be broken until another PR makes it adopt the new toolkit interface. We may want to merge to an intermediate branch and only reach develop when everything is nice and shiny.

packages/toolkit/src/apm/getApmRegistryPackages.js
```diff
@@ -1,24 +1,19 @@
...
  * @param {function(number)} progressHandler Progress handler
+ * @param  {string} environment Envrionment
  * @returns {void}
  */
 export default async (
-  web3,
   apmRegistryName,
-  apmOptions,
-  progressHandler = () => {}
+  progressHandler = () => {},
```
ajsantander: We need to start thinking of a way to improve this. I'd love to move it out of parameters and someway subscribe to it in a completely independent way. So, this function would only accept an environment. Then, you would use an event emitter and an event identifier string to listen to the events in another argument.

ajsantander:
Congrats @0xGabi this is exactly what we were slowly preparing for since we started migrating logic to the toolkit. It's much more usable now. Happy to see so many parameters obliterated!

My review mainly focused on interface changes, which look great. Implementation changes are simply too many for a thorough review. We should be alright though, since we are also aiming for a significant code coverage increase in tests in this sprint.

packages/toolkit/src/apm/getApmRegistryPackages.js
0xGabi: I like that idea and confirm my struggle so far with the `progressHandler`. Would you mind try to include an issue describing a bit more about how that work with events?

PR #1299, opened 2020-01-30
Breaking change to aragonCLI: toolkit interfaces refactor

0xGabi:
# 🦅 Pull Request

Includes #1294, #1236

Refactor middlewares to only use the `environment` option.

The toolkit will be the responsible for setup the specific environment. If the cli run on an Aragon app folder, the toolkit will use the existent `arapp.json` and `truffle.config`. Otherwise we will configure with default settings. 

Add a new pure functionn`configEnvironment` that given an environment option (e.g. `rinkeby`) return an environment object with all the information we need:
```
environment = {
	appName,
	apmOptions,
    provider,
    wsProvider,
    gasPrice,
}
``` 

Remove cli options:  
- `ens-registry`
- `ipfs-rpc`
- `ipfs-gateway`
- `gas-price`


packages/cli/src/cli.js
```diff
@@ -38,14 +35,6 @@ export function init(cb) {
     /**
      * OPTIONS
      */
-    .option('gas-price', {
```
0xGabi: Removes cli global options. Looking forward on the power-cli the powerusers are not going to customize that much their environment and are going to mostly interact with the default environments. 

A future step is adding a new command to setup new environments once. Afterwards those will be available and the toolkit will load them automatically for future requests.

packages/cli/src/middleware/index.js
```diff
@@ -1,60 +1,24 @@
-import { getTruffleConfig } from '../helpers/truffle-config'
```
0xGabi: The middleware logic to setup an environment is not available on aragonCLI anymore and is all handle by the toolkit

packages/cli/src/middleware/index.js
```diff
@@ -1,60 +1,24 @@
...
-  NoNetworkInTruffleConfig,
-} from '../lib/environment/configEnvironment'
+import { loadArappFile } from '@aragon/toolkit'
 
 class InvalidArguments extends Error {}
 
-export function configCliMiddleware(argv) {
-  const [cmd, subcmd] = argv._
-
+export function configEnvironmentMiddleware(argv) {
```
0xGabi: The only transformation that we do on the middleware is parsing the global options `--use-frame` and `--environment` to return a modified environment with: `frame:<environment>` if the `use-frame` was enable.

packages/cli/src/lib/environment/configEnvironment.ts
```diff
@@ -1,203 +0,0 @@
-const Web3 = require('web3')
```
0xGabi: All environment logic is now handle by the toolkit.

packages/cli/src/commands/dao_cmds/new.js
```diff
@@ -56,21 +54,19 @@ export const builder = yargs => {
 
 // Task will be moved to handler once `dao start` is refactored
 export const task = async ({
-  web3,
```
0xGabi: Global options are no longer needed across all aragonCLI commands. Handle on the toolkit.

packages/cli/src/commands/dao_cmds/new.js
```diff
@@ -83,37 +79,26 @@ export const task = async ({
...
-        title: `Fetching template ${bold(template)}@${templateVersion}`,
-        task: async () => {
-          repo = await getApmRepo(web3, template, apmOptions, templateVersion)
-        },
-        enabled: () => !templateInstance,
-      },
       {
         title: 'Create new DAO from template',
         task: async ctx => {
-          daoAddress = await newDao({
```
0xGabi: The interfaces on all requests to the toolkit functions change from requesting options like: `web3`, `apmOptions`, `repo`, `wsProvider`, `gasPrice`. To only ask for an `environmnet` option that will be setup by the toolkit.

packages/toolkit/config/environments.default.json
```diff
@@ -1,8 +1,8 @@
 {
-  "aragon:local": {
```
0xGabi: As the default configuration is the one we fallback when we are not on an aragon project repo, names could be the same.

packages/toolkit/src/apm/useApm.js
```diff
@@ -0,0 +1,14 @@
+import aragonPM from '@aragon/apm'
```
0xGabi: Instead of have a specific function exposing each `apm.js` functionality needed we will expose the whole `apm.js` object.


In future iterations would like to have a similar behavior for `acl` and `kernel`.

packages/toolkit/src/helpers/wrapper.js
```diff
@@ -0,0 +1,96 @@
+import Aragon from '@aragon/wrapper'
```
0xGabi: Rename to wrapper and clean options to hide `@aragon/wrapper`.

packages/toolkit/src/helpers/useEnvironment.js
```diff
@@ -0,0 +1,177 @@
+import Web3 from 'web3'
```
0xGabi: Here we handle the environment logic initialization. We setup web3, apm and ipfs.

packages/toolkit/src/helpers/useEnvironment.js
```diff
@@ -0,0 +1,177 @@
...
+    useFrame
+  )
+
+  // Context
+  if (provider !== prevProvider) {
+    prevProvider = provider
+    web3 = new Web3(provider)
+  }
+
+  return {
```
0xGabi: We have a pure function that return an environment object initialized given a string with the environment key (e.g. rinkeby, mainnet, etc)


Current version of the toolkit is using the `useEnvironmnet` function across all function. A cleaner solution will be to initialize the toolkit once and call `useEnvrionment` during initialization. Afterwards all logic on the toolkit can access the initialized functionality. This change will further improve API interface having something like:
```
import Toolkit from '@aragon/toolkit'

const toolkit = new Toolkit('rinkeby')

const { useApm, daoNew } = tookit
```

packages/toolkit/src/dao/apps.js
```diff
@@ -1,40 +1,37 @@
...
-} from '../helpers/aragonjs-wrapper'
+import { resolveEnsDomain } from '../helpers/resolveAddressOrEnsDomain'
+import { useEnvironment } from '../helpers/useEnvironment'
+import { initWrapper } from '../helpers/wrapper'
 
 /**
  * Return installed apps for a DAO
  *
  * @param {string} dao DAO address
- * @param {Object} options Options
```
0xGabi: Remove the object inputs across all functions and instead the API ask the needed information for each particular funtionality and the envrionment option.

PR #1433, opened 2020-02-12
Update default gas price on environments

### LINKED ISSUE #1102: Update default gas price per network

0xGabi:
# 🦅 Pull Request Description

Fix #1102


ajsantander:
nit: I've found that 20e9 is more communicative than 2000000000 for gas prices. Keeps you from having to squint and count zeroes.

PR #1480, opened 2020-02-17
New folder structure

0xGabi:
# 🦅 Pull Request Description

Move to a new folder structure looking to support granular exports.

## Rational

Segment exports and have sub modules able to import independently.

The following diagram show the new structure:

Screen Shot 2020-02-17 at 10 46 57 AM

In capital letter are the proposed modules.


packages/toolkit/src/apm/apm.js
```diff
@@ -7,7 +7,7 @@ import { useEnvironment } from '../helpers/useEnvironment'
  * @param  {string} environment Envrionment
  * @returns {Object} aragonPM object
  */
-export default async function useApm(environment) {
```
0xGabi: Rename `useApm` to `getApm`

packages/toolkit/config/environments.default.json
```diff
@@ -5,35 +5,17 @@
   },
   "rinkeby": {
     "gasPrice": "15000000001",
-    "apm": {
```
0xGabi: Remove this defaults as they are handle in the `useEnvironment` logic

packages/toolkit/src/dao/new.js
```diff
@@ -28,7 +28,7 @@ export default async function newDao(
   let template
```
macor161: Perhaps we could rename the `new.js` file to `newDao.js` so it reflects the function name.

PR #1494, opened 2020-02-19
WIP: Publish in toolkit

dapplion:
# 🦅 Pull Request Description

This PR implements the publish functionality that should be done by the toolkit, a part of a bigger SDK development flow. It is implemented with these goals:
- Simplest API possible, to be consumed by the buidler-aragon plugin and the CLI
- Purity, leave child_process calls, truffle coupling, etc to the toolkit consumer

**Note, WIP**
- Incomplete tests
- Incomplete `publishRelease` function

# Rationale

**1.** Added necessary helper functions for the publish flow
- Artifact generation
  - getAragonArtifact
  - parseContractFunctions
  - flattenSolidity
- Validation
  - findMissingManifestFiles
  - matchContractRoles
- fs
  - readArtifacts
  - readIgnoreFiles

**2.** Defined a publish flow. See `publish.ts` and its comments. It includes two main functions:
 - `publishRelease`: Given a `dist` path with a built app and built contracts, return an APM transaction object to publish a new version.
 - `validateRelease`: Given a `dist` path ready for publishing, return void if all assets are correct or throw an error.

Both functions accept as first argument the `dist` path of the release and then a config with options, that's it. These functions should either resolve or throw an error. If a front-end wants to ask interactive questions upon errors (such as the CLI does right now) it should call validateRelease and then ask questions with the validation result.

**3.**

*Todo, further work*

Define Ethereum client & IPFS client options passing. How to initialize those clients in a simple way?


0xGabi:
We should reuse this logic on the buidler and cli tools

0xGabi:
After cherry pick the bit that should remind on the toolkit with #1514 
I'll close this one and reuse most of the logic during buidler publish task.

PR #1503, opened 2020-02-23
Refactor APM API and clients

### LINKED ISSUE #1453: Publish usable in node scripts
Be able to use the toolkit in a NodeJS script for example for publishing an Aragon App

dapplion:
# 🦅 Pull Request Description

Fix #1453 

Builds on https://github.com/aragon/aragon-cli/pull/1501. Pursues the following goals

- Move functionality from apm.js to the toolkit
- Migrate it to Typescript
- Uses ethers instead of web3
- Improve and simplify the API
- Delegate uploading content to IPFS to the consumer of the library

This PR preserves `apm.js`  in the toolkit since the CLI is still coupled to the apm.js API and its IPFS uploading functionality. Since the publish functionality of the CLI will be deprecated soon, I didn't modify that part until the way forward is clear.


0xGabi:
After discussion we agree to deprecate `publish` and `run` commands from develop and maintain them on master until v8 land.

PR #1514, opened 2020-02-26
Cherry pick parser and generateArtifact

0xGabi:
# 🦅 Pull Request Description

Cherry pick parser util and artifact generation update from #1494. Finally apply some types corrections.

TODO:

- [x] Fix reminding parser types errors


0xGabi:
@dapplion would you mind take a look to the current type issues on the parser. It seems the `authHelperFunctions` assume the `node.name` is always a string and that's not true as it's an optional argument

PR #1517, opened 2020-02-27
Compile in the build stage and only lint in the lint stage

dapplion:
# 🦅 Pull Request Description

`npm run build` should not succeed if the ts code is not compilable


0xGabi:
Not align with new priorities.

sohkai:
I may be out of the loop here, but this looks like a useful PR that makes developing on the toolkit better?

0xGabi:
Replace by #1563

PR #1542, opened 2020-03-07
Add common initializer argument to APM methods

dapplion:
# 🦅 Pull Request Description

To allow interaction with `buidler-aragon` multi-environment setting for publishing there must be a standardized way of passing an environment between these two libraries.

Current usage is a string "environment"

```js
import { getRepoVersion } from "@aragon/toolkit"

await getRepoVersion("voting", "latest", environment)
```

the proposed change would be used as

```js
import { Apm } from "@aragon/toolkit"

const apm = Apm(providerArg)
await apm.getRepoVersion("voting", "latest")
```

The environment variable will be as flexible as possible accepting:
- A string tag referring to a known network: `"mainnet"`, `"rinkeby"`
- A string URL to be used as JSON RPC endpoint
- An ethers provider instance
- A web3.js provider instance
...etc

## Rational

- Improves usability of the toolkit and centralizes the management of the environment
- Follows the pattern of other common blockchain client tools

About the format of the modules, each submodule exports a setup function that returns the actual functions to be called by the user. This pattern is similar to what is used in [ipfs-http-client](https://github.com/ipfs/js-ipfs/blob/c81c680196f0af82c9dbe7ac1cd404d99e682b1a/packages/ipfs-http-client/src/cat.js#L8).


packages/toolkit/src/setup.ts
```diff
@@ -1,44 +1,180 @@
...
+  function getAragenProvider() {
+    return new ethers.providers.JsonRpcProvider(aragenRpc, {
+      name: 'aragen',
+      chainId: 999999,
+      ensAddress: ensAddress || aragenEnsAddress,
+    })
+  }
+
+  // Default provider for empty strings or undefined
+  if (!providerArg) {
```
0xGabi: I'm not sure if we want to do aragen as default 🤔

dapplion: Me neither but I just respected the current architecture to not have to change that much code

packages/toolkit/src/setup.ts
```diff
@@ -0,0 +1,182 @@
...
+   * Ethers network options allow to provide a custom ensAddress
+   * Most ethers providers accept it so it's defined in the top scope
+   */
+  const networkOptions = ensAddress
+    ? { name: 'custom', chainId: 999, ensAddress }
+    : undefined
+
+  function getProviderFromUrl(url: string): ethers.providers.Provider {
+    const connectionInfo = headers ? { url, headers, timeout } : url
+    return new ethers.providers.JsonRpcProvider(connectionInfo, networkOptions)
```
0xGabi: We might want to use a JsonRpcWebSocketProvider now that's supported by ethers: https://github.com/ethers-io/ethers.js/issues/141

dapplion: But it's on the v5 beta tho

Issue #26, opened 2018-01-09
publish: Add warning if aragon app solidity code initializes any global variable


izqui:
Because of our use of Proxies, child contracts won't initialize global variables when created. In case an app being published does this, we should at least warn the developer.

```.js
contract MyFancyApp is App {
  uint initialState = 1;
}
```

In the above example, when used behind a proxy, `initialState` will be 0, even though the expectation reading the code is that it will be 1.

We should encourage the developer to make it something like: 

```.js
contract MyFancyApp is App {
  uint initialState;

  function initialize() onlyInit { initialState = 1; }
}
```

Edit: this should apply to logic in constructors too

izqui:
This should also make it in the documentation

onbjerg:
Should be fairly simple to do with solidity-parser. Obviously constants are excluded as they're not written to storage.

For the docs I think an illustration would be helpful. Storage *is* initialized, but just not for the proxy, which is the storage we care about.

PR #1543, opened 2020-03-08
26 check global variables

alexmedkex:
# 🦅 Pull Request Description

Addresses issue: https://github.com/aragon/aragon-cli/issues/26

- Add another validation step before generating artifacts, which checks for assignments to global variables in the source code.
- Logs a warning if such assignments were found.

NOTE:

I had to add another parameter which takes the source code of the contract to `getAragonArtifact()`.
I'm not sure if this is the right way to go. Perhaps it would be better to put the app validation into another module separately, which I imagine would be called in `aragon-buidler` after compiling the contracts.

## 🚨 Test instructions

See test files.

## ✔️ PR Todo


alexmedkex:
@0xGabi requesting review :smile:

packages/toolkit/src/utils/appValidation.ts
```diff
@@ -30,7 +34,21 @@ function findFunctionSignatureCollisions(abi1: AbiItem[], abi2: AbiItem[]) {
   return collisions
 }
 
-export function checkSignatureCollisionsWithProxy(abi: AbiItem[]) {
+function checkAssignedGlobalVariables(sourceCode: string) {
+  const assignments = parseGlobalVariableAssignments(sourceCode)
+  const assignments2 = parseConstructorAssignments(sourceCode)
```
macor161: Since the constructor of an app is never called, we could perhaps simply check if the code contains the constructor and output a warning instead of verifying its assignments. 
See https://hack.aragon.org/docs/aragonos-building#constructor-and-initialization for more details about app initialization.

packages/toolkit/src/utils/appValidation.ts
```diff
@@ -30,7 +34,21 @@ function findFunctionSignatureCollisions(abi1: AbiItem[], abi2: AbiItem[]) {
...
+  if (assignments.length > 0 || assignments2.length > 0) {
+    console.log(
```
macor161: I think it would be preferable that the toolkit doesn’t log to the console since it is a library.

packages/toolkit/src/utils/parseContractFunctions.ts
```diff
@@ -192,3 +192,121 @@ export default function parseContractFunctions(
 
   return functions
 }
+
+/**
+ * Finds global storage variable declarations with initialized values, e.g 'int a = 1'.
```
macor161: Added a small clarification :)

```suggestion
 * Finds global storage variable declarations with initialized values, excluding constants. E.g 'int a = 1'.
```

alexmedkex:
Thanks for reviewing! 
Currently, `aragon-buidler` has it's own logic for generating artifacts. I'm wondering if it should use the logic from the toolkit instead...

packages/toolkit/src/utils/appValidation.ts
dapplion: Definitely there should not be a console.log @macor161 said. I would recommend to either throw or return a warnings data structure

packages/toolkit/src/utils/appValidation.ts
alexmedkex: Makes sense. Fixed!

packages/toolkit/src/utils/appValidation.ts
alexmedkex: Added throw declarations.

packages/toolkit/src/utils/parseContractFunctions.ts
alexmedkex: :+1:

PR #1567, opened 2020-03-20
Remove blocking validation

dapplion:
# 🦅 Pull Request Description

Blocker to publish any Aragon App


sohkai:
I think this was due to the validation always failing, because there were collisions with `appId()` and `kernel()`s in the ABIs?

I remember having a conversation about this in another PR or issue; maybe we can just comment this out with a todo comment?

PR #1572, opened 2020-03-20
create-aragon-app: Use new boilerplates

0xGabi:
# 🦅 Pull Request Description

Default to react boilerplate, remove bare boilerplate and remove inquire prompt

### DIFF
--- a/packages/create-aragon-app/src/commands/create-aragon-app.js
+++ b/packages/create-aragon-app/src/commands/create-aragon-app.js
@@ -1,12 +1,12 @@
 const TaskList = require('listr')
 const execa = require('execa')
-const inquirer = require('inquirer')
 const { promisify } = require('util')
 const clone = promisify(require('git-clone'))
 //
 const defaultAPMName = require('../helpers/default-apm')
+const isValidAragonId = require('../helpers/is-valid-aragonid')
 const listrOpts = require('../helpers/listr-options')
-const { installDeps, isValidAragonId } = require('../util')
+const { installDeps } = require('../util')
 const { checkProjectExists, prepareTemplate } = require('../lib')
 
 const templateOptions = {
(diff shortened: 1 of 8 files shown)


packages/create-aragon-app/src/commands/create-aragon-app.js
```diff
@@ -80,34 +70,18 @@ exports.handler = async function({
...
-  }
+  const oldTemplate = template === 'reactWithCli'
 
-  let requiresIPFS = true
-
-  if (template === 'buidler') {
-    requiresIPFS = false
-    console.log(
-      `Warning: You are using the experimental "${template}" boilerplate.`
+  if (template === 'reactWithCli') {
```
macor161: Small suggestion:
```suggestion
  if (oldTemplate) {
```

packages/create-aragon-app/src/commands/create-aragon-app.js
```diff
@@ -180,16 +154,18 @@ exports.handler = async function({
...
 
   return tasks.run().then(() => {
-    reporter.success(`Created new application ${name} in path ./${basename}/
-
-Start your Aragon app by typing:
+    reporter.success(`Created new application ${name} in path ./${basename}/\n`)
 
-  cd ${basename}
-  npm start
+    if (template === 'react')
```
macor161: Why do we only print the info for the buidler boilerplate?

0xGabi: Because in the case of the tutorial we don't want to print it. And in case of the CLI, the steps are more complex than that 😅

0xGabi:
Thank for fixing the tests 🙏

packages/create-aragon-app/src/commands/create-aragon-app.js
macor161: Ah indeed it makes sense not to display it for the tutorial hehe

PR #1578, opened 2020-03-23
Use wsRpc in dao apps command

### LINKED ISSUE #1577: Error: Invalid JSON RPC response on dao apps --all command
## 🐛 Bug Report

The `dao apps -all` command returns the following error when run on mainnet:

```
Invalid JSON RPC response: "<html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n<hr><center>nginx/1.17.6</center>\r\n</body>\r\n</html>\r\n"
```

Using the infura endpoint doesn't seem to solve the problem.

macor161:
# 🦅 Pull Request Description

- Fixes #1577 

## Rational

The wsRpc provider wasn't used to fetch permissionless apps.

### DIFF
--- a/packages/cli/src/commands/dao_cmds/apps.js
+++ b/packages/cli/src/commands/dao_cmds/apps.js
@@ -1,6 +1,7 @@
 import TaskList from 'listr'
 import Table from 'cli-table'
 import { blue, green, white } from 'chalk'
+import Web3 from 'web3'
 import {
   getDaoAddress,
   getInstalledApps,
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/dao_cmds/apps.js
```diff
@@ -115,7 +116,7 @@ export const handler = async function({
         task: async (ctx, task) => {
           appsWithoutPermissions = (
             await getAllApps(daoAddress, {
-              web3,
+              web3: new Web3(wsProvider || web3.currentProvider),
```
0xGabi: Why do we need to initialize a new Web3 object? Shouldn't be the same as line 107 🤔

macor161: Good question. Because the `ensureWeb3()` function that generates the `Web3` object doesn't use `wsProvider`. I think this has been fixed in `develop` but not yet on `master`

PR #1592, opened 2020-03-28
Don't insist on using Infura, recommend running one's own eth node

xuhcc:
# 🦅 Pull Request Description

Changed the recommendations printed by CLI.

## Rational

Ethereum is P2P network, so it is important to remind users that they could run their own nodes instead of using centralized services like Infura.

## 🚨 Test instructions

N/A

## ✔️ PR Todo

### DIFF
--- a/packages/cli/src/middleware/index.js
+++ b/packages/cli/src/middleware/index.js
@@ -61,7 +61,7 @@ export function configCliMiddleware(argv) {
     if (wsProvider?.connection?.url?.includes('eth.aragon.network')) {
       reporter.newLine()
       reporter.warning(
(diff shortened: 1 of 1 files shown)


sohkai:
I agree in principle with this, but most users will not want to sync a local geth node or buy a dappnode just so they can use this CLI 😄.

I think we could print out an additional message to compliment the Infura suggestion though.

xuhcc:
Regardless of what majority would do, I think it is important to communicate this message. Afterall, most users are totally fine with traditional services and organizational structures, yet we're still building the web3 tech.

So, what wording would you suggest?

xuhcc:
Thanks!

PR #1602, opened 2020-04-03
Feat: Update create-aragon-app to support yarn installation on new bo…

0xGabi:
TODO:

- [x] Update docs mention yarn

- [x] Sanity check PATH to know if yarn binary is available otherwise default to npm

### DIFF
--- a/packages/create-aragon-app/README.md
+++ b/packages/create-aragon-app/README.md
@@ -26,7 +26,7 @@ Once finished, this will open [http://localhost:3000/](http://localhost:3000/) i
 
 _Note: [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b) comes with npm 5.2+ and higher. If you use npm 5.1 or earlier, you can't use `npx`. Instead, install `create-aragon-app` globally._
 
-_Note: We also support [yarn](https://yarnpkg.com/getting-started/install) as a package manager during installation. Make sure to have it on your PATH._
+_Note: We also support [yarn](https://yarnpkg.com/getting-started/install) as a package manager during installation and will use it by default if you have it available in your PATH._
 
 # Tests
 
--- a/packages/create-aragon-app/README.md
+++ b/packages/create-aragon-app/README.md
@@ -26,6 +26,8 @@ Once finished, this will open [http://localhost:3000/](http://localhost:3000/) i
 
 _Note: [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b) comes with npm 5.2+ and higher. If you use npm 5.1 or earlier, you can't use `npx`. Instead, install `create-aragon-app` globally._
 
+_Note: We also support [yarn](https://yarnpkg.com/getting-started/install) as a package manager during installation. Make sure to have it on your PATH._
+
 # Tests
 
 In the root of the repository, run:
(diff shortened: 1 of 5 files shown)


packages/create-aragon-app/src/util.js
```diff
@@ -1,10 +1,14 @@
+const commandExistsSync = require('command-exists').sync
 const execa = require('execa')
 
 const PGK_MANAGER_BIN_NPM = 'npm'
 const PGK_MANAGER_BIN_YARN = 'yarn'
 
 const getNodePackageManager = oldTemplate => {
-  return oldTemplate ? PGK_MANAGER_BIN_NPM : PGK_MANAGER_BIN_YARN
+  if (commandExistsSync('yarn')) {
```
sohkai, line 8: This might be simpler as:

```js
if (commandExistsSync('yarn') && !oldTemplate) {
  return PGK_MANAGER_BIN_YARN
}

return PGK_MANAGER_BIN_NPM
```

packages/create-aragon-app/README.md
```diff
@@ -26,6 +26,8 @@ Once finished, this will open [http://localhost:3000/](http://localhost:3000/) i
 
 _Note: [npx](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b) comes with npm 5.2+ and higher. If you use npm 5.1 or earlier, you can't use `npx`. Instead, install `create-aragon-app` globally._
 
+_Note: We also support [yarn](https://yarnpkg.com/getting-started/install) as a package manager during installation. Make sure to have it on your PATH._
```
sohkai, line 29:
```suggestion
_Note: We also support [yarn](https://yarnpkg.com/getting-started/install) as a package manager during installation and will use it by default if you have it available in your PATH._
```

packages/create-aragon-app/src/util.js
```diff
@@ -1,13 +1,21 @@
...
-const getNodePackageManager = () => {
-  return PGK_MANAGER_BIN_NPM
-}
+const getNodePackageManager = useYarn =>
+  useYarn ? PGK_MANAGER_BIN_YARN : PGK_MANAGER_BIN_NPM
+
+const installDeps = (oldTemplate, cwd, task) => {
+  const useYarn = commandExistsSync('yarn') && !oldTemplate
+
+  // If we don use yarn we delete yarn.lock
```
sohkai, line 15:
```suggestion
  // If we don't use yarn we delete yarn.lock
```

packages/create-aragon-app/src/util.js
```diff
@@ -1,13 +1,21 @@
...
+  if (!useYarn) fs.removeSync(path.join(cwd, 'yarn.lock'))
```
sohkai, line 16: I think there's also the `app/yarn.lock` file.

macor161: Actually, I would perhaps keep the `yarn.lock` files. It's not necessarily bad to support both yarn and npm.

packages/create-aragon-app/src/commands/create-aragon-app.js
```diff
@@ -160,7 +161,7 @@ exports.handler = async function({
       reporter.info(`Start your Aragon app by typing:
 
     cd ${basename}
-    npm start
+    ${commandExistsSync('yarn') ? 'yarn' : 'npm'} start
```
0xGabi, line 164: Good catch 💪

PR #1616, opened 2020-04-07
Update dependencies

macor161:
# 🦅 Pull Request Description

- Bump `prettier` to `v2.0.4` and fix lint
- Bump `ipfs-http-client` to `v41.0.1` and fix missing packages. `v43.0.0` and `v42.0.0` were causing a large number of tests to fail.
- Update stable releases in readme
- Closes #1615


packages/toolkit/package.json
```diff
@@ -46,7 +46,8 @@
     "fs-extra": "^9.0.0",
     "get-folder-size": "^2.0.1",
     "go-platform": "^1.0.0",
-    "ipfs-http-client": "^40.1.0",
+    "ipfs-http-client": "^41.0.1",
```
sohkai, line 49: Wondering—do you think we should investigate why the new versions were failing, or should we just soft pin to this version using `~`?

packages/cli/package.json
```diff
@@ -198,6 +198,7 @@
   "prettier": {
     "singleQuote": true,
     "semi": false,
+    "arrowParens": "avoid",
```
sohkai, line 201: Haha actually... could we leave this as its default? We will most likely be using the same settings in the other repos :).

macor161: Ah yes that was a difficult one to choose hehe. No strong opinion for or against the rule but I found the parenthesis around all the `t` in our tests not to be terribly necessary. However, if we're using the same rule in other repos, I added it back for consistency.

packages/toolkit/package.json
macor161: I think it’s definitely worth investigating further. `ipfs-http-client` is a relatively popular and well-maintained library so this could point to a problem in our code.

PR #1639, opened 2020-04-20
Fix CI install dependencies failure

macor161:
# 🦅 Pull Request Description

- Use `package-lock.json` instead of `npm-shrinkwrap.json`
- Pushed a few empty commits to test the CI multiple times

## Rational

It's still unknown why the Github actions fail randomly with `npm-shrinkwrap.json` files but they seem to be much more stable with `package-lock.json`.


sohkai:
We should only be including shrinkwraps for the CLI and `create-aragon-app`.

[From the documentation](https://docs.npmjs.com/files/shrinkwrap.json):

> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It’s strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

Shrinkwraps are still helpful to include when we publish, but perhaps we can only [generate them ahead of publishing](https://docs.npmjs.com/cli/shrinkwrap) rather than always having them committed (we should favour `package-lock.json` when developing / testing)?

macor161:
> Shrinkwraps are still helpful to include when we publish, but perhaps we can only generate them ahead of publishing rather than always having them committed
Ah I didn't think about this solution, that's a good idea!

macor161:
> I guess we can move to yarn.locks in a different PR?

Actually, I thought it wouldn't be too difficult to add the yarn lockfiles to this PR but I stumbled upon errors when using `yarn import`. So I created [this issue](https://github.com/aragon/aragon-cli/issues/1647) for hopefully next week.

PR #1662, opened 2020-05-09
Add throttle to wrapper.apps observable

### LINKED ISSUE #1661: Transaction path not found
## 🐛 Bug Report

The transaction path is not found sometimes when executing various commands like `dao exec`, `dao install` or `dao acl`. The probable cause is that we don't wait long enough to fetch all the permissions. 

A possible fix could be to add a throttle of a few seconds to the observable.

macor161:
# 🦅 Pull Request Description

- Fixes #1661 
- Adds a throttle of 3 seconds.

WIP, was not able to replicate #1661 yet.


macor161:
Still unable to replicate the issue. I will merge the PR and we'll see if we keep receiving bug reports about this.

PR #1686, opened 2020-05-19
Reorganize ganache dependencies

macor161:
# 🦅 Pull Request Description

- Fix missing `ganache-core` dependency
- Remove unused `ganache-cli` dependency

## Rational

Seems like moving to yarn uncovered a problem in our dependencies. Trying to run `@aragon/cli/7.1.4` throws a `Cannot find module 'ganache-core'` error.

`ganache-core` was moved  to the dev dependencies since `v7.1.0` but it is a production dependency. I moved it back to the `dependencies` section. `ganache-cli` is never used so it was removed.


macor161:
> LGTM! The aragen/devchain commands only use ganache-core?

Yep only `ganache-core` here: https://github.com/aragon/aragon-cli/blob/master/packages/cli/src/commands/devchain_cmds/start.js#L13

PR #1687, opened 2020-05-20
Generate artifact's signatures from ABI when possible

### LINKED ISSUE #1673: Wrong artifact generation with enums
## 🐛 Bug Report

- The artifact generator converts enums to `address` instead of `uint..`.

## To Reproduce

1. Clone https://github.com/aragonone/voting-connectors.git
2. Move into the `voting-aggregator` app folder
3. Run `aragon apm publish major --artifact-only`
4. See `addPowerSource(address,address,uint256)` instead of `addPowerSource(address,uint8,uint256)`

macor161:
# 🦅 Pull Request Description

- Fixes #1673  

## Rational

The artifact generator previously converted any unknown types (e.g. `PowerSourceType`) to `address`, which works for contracts but not enums. In this PR, we attempt to get ABI nodes matching the function name and number of parameters. If a single node is found, it generates the signature with it, otherwise it still uses the manual logic. Generating the artifact from the AST instead of the source would be even better but this should at least fix the enum signatures in most cases.

## 🚨 Test instructions

`aragon apm publish major --artifact-only` in voting-aggregator app.

## ✔️ PR Todo

### DIFF
--- a/packages/cli/src/commands/extract-functions.js
+++ b/packages/cli/src/commands/extract-functions.js
@@ -2,6 +2,7 @@ import path from 'path'
 import { blue } from 'chalk'
 import TaskList from 'listr'
 import { extractContractInfoToFile } from '@aragon/toolkit'
+import { exists, readJson } from 'fs-extra'
 
 export const command = 'extract-functions [contract]'
 export const describe = 'Extract function information from a Solidity file'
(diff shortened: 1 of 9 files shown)


macor161:
@sohkai hmm good point. Fortunately we are deprecating these commands pretty soon.

I'm not entirely sure the buidler plugin has the same problem but I will verify tomorrow.

PR #1708, opened 2020-06-02
Add missing gasPrice parameter

macor161:
# 🦅 Pull Request Description

Adds missing `gasPrice` parameter to:

- `dao act`
- `dao new`
- `dao exec`


macor161:
So after reverting the last dependency updates one by one, it turns out the `ipfs-http-client` was the problem. Still not sure why tests from [the PR](https://github.com/aragon/aragon-cli/pull/1707) were not failing.

sohkai:
Interesting, perhaps we can have a look at the release notes of that package. We were having an issue fetching the artifacts off of IPFS in the tests, right?

macor161:
> We were having an issue fetching the artifacts off of IPFS in the tests, right?

Mostly yes, but sometimes other unrelated tests as well.

> Maybe v44 requires IPFS 0.5/0.6 or something?

Ah maybe! I will look into this.

PR #1731, opened 2020-06-21
Use Aragon Connect in dao apps command

macor161:
# 🦅 Pull Request Description

- Use Aragon Connect to fetch apps when possible (i.e. on mainnet and Rinkeby)

## 🚨 Test instructions

`dao apps <DAO address> --env aragon:rinkeby`

### DIFF
--- a/packages/cli/src/commands/dao_cmds/apps.js
+++ b/packages/cli/src/commands/dao_cmds/apps.js
@@ -92,7 +92,7 @@ const supportsAragonConnect = (networkId) => {
 }
 
 const getAppsFromAragonConnect = async (dao, networkId) => {
-  const org = await connect(dao, ['thegraph'], { chainId: networkId })
+  const org = await connect(dao, 'thegraph', { chainId: networkId })
 
   return (await org.apps()).map((app) => ({
     appId: app.appId,
--- a/packages/cli/src/commands/dao_cmds/apps.js
+++ b/packages/cli/src/commands/dao_cmds/apps.js
@@ -1,6 +1,7 @@
 import TaskList from 'listr'
 import Table from 'cli-table'
 import { blue, green, white } from 'chalk'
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/dao_cmds/apps.js
```diff
@@ -79,6 +82,27 @@ const printPermissionlessApps = (apps) => {
   }
 }
 
+/**
+ * Returns true if the provided Ethereum network id supports
+ * Aragon Connect. Currently `true` for mainnet and Rinkeby
+ * @param {number} Ethereum network id
+ */
+const supportsAragonConnect = (networkId) => {
+  return networkId === 1 || networkId === 4
```
sohkai, line 91: Thinking aloud here: the only reason we would want to continue using aragon.js going into the future is for local development environments (e.g. with buidler), because we wouldn't want to force them to run a graph node locally?

This is a compelling reason for us to eventually build that ethereum connector that we can fallback to if a subgraph is not available / does not contain this org (for whatever reason).

macor161: Yep we can likely remove aragon.js once we stop supporting the local devchain in aragonCLI.

0xGabi: >because we wouldn't want to force them to run a graph node locally

@sohkai you are right it is a good reason. On the other hand, we realize working on a couple of Aragon apps subgraphs that it was esier to a developer using The Graph playground infrastructure directly and just have a staging subgraph that using the graph node locally. I think the main downside of this approach is relying heavily on The Graph infrastructure, but we were able to go a long way with it so far.

sohkai: Right, the problem with this (for the buidler environment), is that we most likely don't want the user to run their own subgraph infrastructure for their local chain.

PR #1737, opened 2020-06-22
Client hotfix

macor161:
# 🦅 Pull Request Description

- Related to https://github.com/aragon/aragon/pull/1434
- Uses a temporary client at https://github.com/0xGabi/aragon until the client bug is fixed
- Increases the devchain timeout to 5m to fix CI setup problem


macor161:
> Can't believe it takes more than 1.5min to start up the devchain now...

Well only in the CI fortunately.

Issue #1764, opened 2020-07-14
'dao install' sent a transaction which doesn't install an app


ethene:
## 🐛 Bug Report

After executing 'dao install' the tool has executed a transaction which doesn't end up with the app being installed

### Have you read the [Contributing Guidelines on issues](https://github.com/aragon/aragon-cli/blob/master/CONTRIBUTING.md#ways-to-contribute)?

- [x] Yes

## To Reproduce

aragon dao install stablefund.aragonid.eth token-request.aragonpm.eth --app-init-args 0xe8aa182d10252bed21cbe35a1f8cac9b0d9458a7 0xc1ee97ce717c1dfed21543ca56f23303c1d5dffc "[]" --environment aragon:mainnet --gas-price 37
✔ Fetching token-request.aragonpm.eth@latest
✔ Fetching token-request.aragonpm.eth@latest
↓ Checking installed version [skipped]
→ Installing the first instance of token-request.aragonpm.eth in DAO
✔ Deploying app instance
↓ Fetching deployed app [skipped]
→ App wasn't deployed in transaction.

ℹ Successfully executed: "Execute desired action as a token holder"

TX:
https://etherscan.io/tx/0x11911c9abc21ab2f9545088276fcfbca6a237a3328a377e335d3ddc105194092

App is not seen in the configuration, on UI or 'aragon dao apps'

## Expected behavior

I expect the cli to install the new app which can be seen and configured

## Actual Behavior

TX created:
https://etherscan.io/tx/0x11911c9abc21ab2f9545088276fcfbca6a237a3328a377e335d3ddc105194092

app not seen anywhere in the configuration

## Context

mainnet

#### Organization

stablefund.aragonid.eth

#### Environment

MacOS Mojave
CLI from NPM

0xGabi:
Hi @ethene this an expected behavior. If you look into the success message: `Execute desired action as a token holder` what the aragonCLI did was trying to install the app but as you don't have the full permissions to do it directly it creates a Vote that if gets approve the app will be installed. Looking into your organization, this is the open vote: https://mainnet.aragon.org/#/stablefund/0x241b27133d78177f1fb37ea26645bff0d1e1bbad/vote/0/

If you are the only owner fo the organization and you would like to do a quick setup without having to vote on every change, you can assign your account the role to create permissions: https://mainnet.aragon.org/#/stablefund/permissions/app/0xe0f35ac8e6c8b8ab88d4798becc0865496d623ba

[...]

PR #1848, opened 2021-03-30
ARA-667 'dao install'

c0rv0s:
# 🦅 Pull Request Description

Original issue: #1764

In the bug report the user describes an issue with installing an app and having it not show up in the DAO UI. However they didn't realize that the app installation process had to be completed with a vote.

The changes here will check if a vote is required to finish installation of an app and inform the user once the install command finishes executing.

A follow up issue ARA-822 has been added to add a test case for this.

## Rational
The app install process is functioning as intended but the communication to the user of the next steps required was insufficient.

## 🚨 Test instructions

Install a new app into a dao where the Voting app has the APP_MANAGER permission and at the end of the install command there will be a message saying to check the Voting app for a new vote to complete the installation.


nestorbe (approved):
LGTM!

cbrzn (approved):

PR #1892, opened 2021-09-28
DAO-338 add polygon config

yuetloo:
With this change, we can run the command line without the arapp.json file in the folder where the command is run and would not get the error `truffle.js` not found.


yuetloo:
Close this PR as the fix done in Aragon Client, so, we no longer need this command line change

PR #1962, opened 2024-01-08
Update README.md

AaronAbuUsama:
Add deprecation notice

## Rational

This repo is causing confusion


carlosgj94 (approved):
Agreed. 
This repo is not only useless and doesn't work, but also creates confusion. Let's archive it as well after passing this PR

